PR #11, opened 2018-09-14
Clean branch

JayBuckley7:
Now this one I'm sure is a security error that's just going to be different at every company so by no means am I saying there is a problem or that I have the solution. When attempting to go out and get the updated PSL something breaks probably on my end and i get "failed to update PSL  -------401 request

I added a method in Cli that makes the download go through for me, but it may not be the best way

IIn elastic and init I've noticed the way the methods seem to work is take all the emails in the box, parse them into xml, add them to a huge index in elastic

In trying this myself I'd usually find find one or two bad emails in the middle that would raise an exception like no source_IP then the whole program would stop and it wouldn't index any of the mail unless you pulled that box, so 2 methods in INIT that pull the emails one at a time then parse then individually and failover and skip if they get a bad email, its less efficient but its been working for me. in doing this i also surrounded the relevant methods in elasitc.py in try catch methods and changed their raises to print.

These methods were mostly ment for my use case and they just assume the user wants the searches indexed with elastic so they may not be the best for general use cases


JayBuckley7:
the "clean-branch" is because my master is way to out of sync with yours lol I just forced your current branch on top of mine on a new path :)

seanthegeek:
Hi @JayBuckley7,

 I fixed the crashing on forensic reports earlier, but it doesn't look like it made its way into your fork. Also, when you  have some time, try out the `4.0` branch and see if it fixes the PSL issue too.

Use this command to upgrade to this development branch:

```bash
sudo -H pip3 install -U git+https://github.com/domainaware/parsedmarc.git@4.0
```

Then restart the `parsedmarc` service:

```bash
sudo service parsedmarc restart
```

`parsedmarc` 4.0 will be released as soon as I finis adding Splunk support.

JayBuckley7:
yep! with 4.0 I can use the production build with no changes!

PR #13, opened 2018-09-25
Smtp auth

SergioFernandezCordero:
- Added parameters to specify if SMTP will use SSL-TLS or not and its port
- Changed the way the function email_results is invoked to reflect this cases
- Changed the function email_results to support SSL, TLS or plain connections
- Changed the way email_results connects to SMTP servers to comply with Gmail methodology


seanthegeek:
Thanks! I just implemented a variation of your fix. It will always try to use STARTTLS if possible.

Give the latest build a try.

PR #21, opened 2018-10-10
Add Kafka Support

mikesiegel:
Hello!

This change is to introduce the ability to output to a Kafka topic.  This is useful in our case when an ELK is under heavy load, Kafka topics are used as queuing mechanism prior to Logstash.  Maybe others will find this useful a well? It's pretty straight forward and I tried to stay similar to the style of the rest of the code base (PEP8/function naming etc).


seanthegeek:
Thsnks

PR #50, opened 2019-02-06
fixes elasticsearch output

andrewmcgilvray:
There were many problems created by what looks like a bad search and replace.


seanthegeek:
D'oh! Thanks!

PR #62, opened 2019-02-15
Add parallelized processing option for parsing reports with the CLI

zscholl:
# Summary
I added two parameters to the config file:
`n_cpus`
`chunksize`

if `n_cpus` is set to a number higher than 1, then when parsing through the CLI parsedmarc will start as my processes as is specified in `n_cpus`.

`chunksize` is the number of files each process will be given to process at a time. When there are thousands of files to parse there is extra overhead in passing additional files to a new process so setting this to higher than one can increase throughput.

This change will disable geo_lookups with the maxmind database when `n_cpus` is set to a number higher than one. There were some issues accessing the maxmind database concurrently and at the time we did not need the geo-lookup data so the easiest solution was to disable it when operating in parallel mode.

I also added a progress bar with tqdm. This is nice if you have tens or hundreds of thousands of files to parse.

# Performance
I haven't yet done any rigorous testing, but with some crude estimation we went from ~6 sec/file to 6 files/sec. I was able to kick off 400 processes on a machine with 2 Xeon E5-2640s, CPU usage barely broke 2% and memory usage crept up to 18gigs after 2 hours of running. 

There are a lot of efficiencies to be gained yet, but this helped us go from needing months to parse a set of 300k reports to being able to get it done in a weekend.


zscholl:
Pulled in master, updated `n_cpus` to `n_procs`, fixed pep8.

Tests.py was hung because of an erroneous `pdb.set_trace()` I left in `utils.py` it has now been removed.

Still need to do some testing on the geo_ip bit, it should be able to handle concurrent read access, so there may be some other issue. 

I have not explicitly tested IMAP but the multiprocessing occurs in a separate branch of code. It will only be engaged when passing a filepath through the CLI, all other code paths have the same behavior they used to.

And looks like I need to update the Namespace parameters.

seanthegeek:
Awesome. Thanks!

seanthegeek:
@zscholl Did I break your code? Please see issue #147

zscholl:
> @zscholl Did I break your code? Please see issue #147

I just ran a test on the samples in the repo and it worked as expected given the config options in the issue. Hard to say without some more information

PR #98, opened 2019-07-18
Expose CSV rows as a list of dicts

michaeldavie:
Expose the flattened records as a list of dicts, so that they can be modified by an external script before being written to a CSV.


seanthegeek:
Thanks

michaeldavie:
Awesome. Between this and the offline mode I should be able to use the library as is. Thanks for all your work on this!

PR #100, opened 2019-07-25
Correct bug in 6.5.1

michaeldavie:
This change corrects a bug introduced in 6.5.1 that caused only the last record's data to be used for each row in an aggregate report's CSV version.


seanthegeek:
Thanks. I'll get this into a release shortly.

PR #119, opened 2019-10-27
Fix typos

ruffy91:
Fixed some typos.
Also fixes a typo in the elasticsearch fields.
You can update the field for existing data using a query like this:
```
POST dmarc_aggregate-*/_update_by_query
{
    "query": { 
    "exists": {
      "field": "xml_schemea"
    }
  },
    "script" : {
        "inline": "ctx._source.xml_schema = ctx._source.xml_schemea; ctx._source.remove(\"xml_schemea\");"
    }
}
```


seanthegeek:
@ruffy91 D`oh! Thanks for the fixes. Could you add the changes to `kibana/export.ndjson` as well?

PR #122, opened 2019-11-06
Handle invalid aggregate reports

michaeldavie:
In my testing over the past few months I've encountered some invalid aggregate report files, and I'd like to upstream my workarounds so that the errors can be handled gracefully rather than discarding an entire report due to a single bad character.

Specifically, I've observed the following two errors:

* Invalid bytes (such as 0x91), typically in URL fields.
* Invalid XML, typically due to the use of `<` or `>` characters in user-defined fields (e.g. org email) or in a URL.

I've added sample report files that exhibit these characteristics, and have made the modifications needed to handle them, including:

* Dropping invalid bytes when decoding
* Parsing invalid XML with `lxml` (a new dependency) and attempting to recover from errors
* Stripping out any artifacts introduced by `lxml` as part of its recovery efforts

I also made one modification at line 422/429 of `__init__.py` to correct what may have been a bug.

I'm not sure if this is a perfect solution, so please test it out and let me know what you think.

### DIFF
--- a/parsedmarc/__init__.py
+++ b/parsedmarc/__init__.py
@@ -24,6 +24,7 @@ import mailbox
 import mailparser
 from expiringdict import ExpiringDict
 import xmltodict
+from lxml import etree
 from mailsuite.imap import IMAPClient
 from mailsuite.smtp import send_email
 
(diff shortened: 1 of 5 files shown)


seanthegeek:
Thanks for these improvements, I'd still like to have a way of letting people that these are invalid/bad reports, so can you add an error message to the errors list in the existing `report_metadata["errors"]` list in the results for each workaround?

Also, I have no idea why TravisCI is saying lxml 4.4.1 does not exist.

https://travis-ci.community/t/pip-cannot-find-lxml-4-4-1-even-though-it-exists/5807

For now, can you set the dependency to lxml>=4.4.0?

michaeldavie:
Looks like you beat me to it. Thanks for merging this in!

PR #132, opened 2019-12-17
Making dkim_aligned and spf_aligned case insensitive.

aharpour:
Hi 
First of all I would like to thank you for making parsedmarc an open source project. 
I ran parsedmarc on our company dmarc reports. When I was sorting through the results I notice all the report from one particular email provider has their “dkim_aligned”, “spf_aligned” and “passed_dmarc” set to false. But when I look at the reports I could see that both spf and dkim are aligned. After a bit of debugging it turned out that in the report instead of “pass” it is written “Pass” and the code checks pass in a case sensitive way.

### DIFF
--- a/parsedmarc/__init__.py
+++ b/parsedmarc/__init__.py
@@ -115,8 +115,8 @@ def _parse_report_record(record, offline=False, nameservers=None,
     if "spf" in policy_evaluated:
         new_policy_evaluated["spf"] = policy_evaluated["spf"]
     reasons = []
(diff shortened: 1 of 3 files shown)


seanthegeek:
Good catch. Thanks!

seanthegeek:
I just released parsedmarc 6.7.3, which includes this PR. Thanks again, @aharpour!

aharpour:
@seanthegeek  Wow! already released that is awesome . Thanks

PR #135, opened 2019-12-23
Peg geoip version to 2.9.0 to avoid dep version conflict

### LINKED ISSUE #134: urllib version conflict
When installing I get:
```
 ERROR: geoip2 3.0.0 has requirement urllib3>=1.25.2, 
 but you'll have urllib3 1.24.3 which is incompatible.
```

This is because the `geoip2>=2.9.0` requirement is now pulling geoip2 3.0.0 which requires `urllib3>=1.25.2`. But this package requires `urllib3<1.25,>=1.21.1` which is a conflict.

### Proposed solution

Set this package to require `geoip2==2.9.0` exactly.

lukecyca:
Also fixed pep8 formatting of dependencies in `setup.py`.

fixes #134

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -7,7 +7,7 @@ urllib3<1.25,>=1.21.1
 requests>=2.2.16.0
 publicsuffix2>=2.20190812
 xmltodict>=0.12.0
-geoip2>=2.9.0
+geoip2==2.9.0
 imapclient>=2.1.0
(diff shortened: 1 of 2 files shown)


seanthegeek:
I just updated all of the related dependencies instead, so I wouldn't need to maintain old versions. Thanks though!

PR #146, opened 2020-02-13
Remove nameserver warning

Wouter0100:
This warning seems a bit.. unneeded. I would say this is up to the people that implement the software and not to the developer.

We have been monitoring both Google DNS and Cloudflare DNS, and concluded that from our AS - Google DNS is way more reliable with a lower and more consistent latency. We have had survival cases where Cloudflare DNS did not work properly and showed weird latency spikes, while Google was constantly performant.

Due backwards compatibility I was unfortunately not able to remove the defaulted nameservers to Cloudflare, but I'd suggest considering this in a BC-breaking change and default to the system's NS.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -192,17 +192,6 @@ The full set of configuration options are:
     - ``attachment`` - str: The ZIP attachment filenames
     - ``message`` - str: The email message (Default: Please see the attached parsedmarc report.)
 
-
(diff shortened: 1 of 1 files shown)


seanthegeek:
I'll remove the default nameservers in a future release.

PR #151, opened 2020-03-11
Use SMTP port from config when sending the report

Yabk:
The SMTP port from config isn't being used when sending the report.
Instead, `send_email` function from `mailsuite.smtp` is called with `port=0` which causes that the default port 465 is always used regardless of the config.

Had an error in previous PR so I made a new one. This time I made sure to test things out before pushing.


seanthegeek:
Thanks!

PR #157, opened 2020-04-07
Fix for copy-paste error

sander85:
If max_retries is defined in config it will override the port configuration.

Signed-off-by: Sander Lepik <sander.lepik@cooppank.ee>


seanthegeek:
Thanks!

PR #160, opened 2020-05-06
fix: [splunk] correct default index for splunk dashboard

cvandeplas:
This PR fixes the default index of the XML which was set to `email_ess` instead of the `email` which is used in the rest of the code and the aggregate dashboard. This change therefore makes it more consistent.


seanthegeek:
Good catch. Thanks!

PR #170, opened 2020-06-10
Fixing forensic parse failure on valid forensic report

danhstevens:
Many of our Forensic Reports include the following in the feedback report:

```
Source-IP: 54.240.8.61 (a8-61.smtp-out.amazonses.com)
```

Since the reverse DNS is included inside the parenthesis instead of only an IP address, calling `get_ip_address_info` fails, which in turn causes the entire forensic report to fail as invalid.

This PR simply strips out anything after the IP address so the forensic report can be processed correctly.


nhairs-finder:
@seanthegeek is there any timeline for uploading a new release to PyPi? All my forensic reports are impacted by this bug 😞 

I'm happy to provide assistance if needed 🙂

seanthegeek:
@nhairs-finder This, and a bunch of other fixes were just released in 6.11.0 😄

PR #191, opened 2020-10-19
Allow Basic Auth with Elasticsearch

knightian:
Hello,

Elasticsearch lets you use the xpack.security to set passwords for free now.

Everybody should use passwords, and so I have made it such that it no longer gives me 401 when I enabled a password on Elasticsearch.


seanthegeek:
Thanks!

Issue #197, opened 2020-11-05
parsedmarc stoops working if file name of sample mail is too long


arne1101:
If sample mail has a too long filename, parsedmarc crashed with 
```
File "/usr/local/bin/parsedmarc", line 8, in <module>
    sys.exit(_main())
  File "/usr/local/lib/python3.7/site-packages/parsedmarc/cli.py", line 612, in _main
    save_output(results, output_directory=opts.output)
  File "/usr/local/lib/python3.7/site-packages/parsedmarc/__init__.py", line 1296, in save_output
    with open(path, "w", newline="\n", encoding="utf-8") as sample_file:
OSError: [Errno 36] File name too long ...
```

and then stopps. I would suggest skipping that particular file and proceed with the next or truncate the filename before saving that file.

PHPGangsta:
I'm not a Python expert, but my feeling is that the function get_filename_safe_string() is not a good idea.

https://github.com/domainaware/parsedmarc/blob/82b1a7e29243a9ca6015e078600612ae3e6ba5ad/parsedmarc/utils.py#L381

What happens if you put ".." into this function? The result is an empty string. Not sure what happens with an empty string when trying to write to an empty filename here:
https://github.com/domainaware/parsedmarc/blob/82b1a7e29243a9ca6015e078600612ae3e6ba5ad/parsedmarc/__init__.py#L1296

As @arne1101 says, the length is another problem, operating systems have different filename-length or path-length restrictions.

Only a few invalid characters are removed:
https://github.com/domainaware/parsedmarc/blob/82b1a7e29243a9ca6015e078600612ae3e6ba5ad/parsedmarc/utils.py#L390-L391
What happens to horizontal tabs \t, vertical tabs, non-breaking-spaces, or control codes?

I would suggest to use a character-whitelist, not a character-blacklist. Only allow "a-zA-z0-9_-" for example, remove everything else. If it's empty after the removal of all "not-whitelisted characters", set it to "None". If it's longer than 30 characters, cut it to 30.

arne1101:
I created a pull request to limit the filename length to 255 characters. It'll just truncate after 255 chars.

PHPGangsta:
parsedmarc might run on Windows. Windows has a limit of the **whole path** of length 260 characters:

"In editions of Windows before Windows 10 version 1607, the maximum length for a path is MAX_PATH, which is defined as 260 characters"
https://docs.microsoft.com/en-us/windows/win32/fileio/naming-a-file

[...]

PR #199, opened 2020-11-06
Limit filename length to 100 characters

arne1101:
https://github.com/domainaware/parsedmarc/issues/197

Since there is a limit for filename length in some OS, filename length should be limited. I propose limit it to 255 characters.


arne1101:
Further reduces to 100 as suggested in https://github.com/domainaware/parsedmarc/issues/197#issuecomment-723039596

PR #222, opened 2021-02-05
Add option to process messages in batches

tom-henderson:
When processing a large number of messages over IMAP, if there is a crash before `process_reports()` is called, nothing will be saved and all the messages will need to be reprocessed. By setting `batch_size` we can save incremental progress.


tom-henderson:
@seanthegeek Any chance you could consider merging this?

PR #223, opened 2021-02-05
Allow saving reports to S3

tom-henderson:
Allows reports to be saved to an S3 bucket. Each report is stored as a single json file under a `year=xxxx/month=xx/day=xx` key path suitable for querying with Athena / QuickSight.

### DIFF
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -132,11 +132,15 @@ For example
    token = HECTokenGoesHere
    index = email
 
+   [s3]
+   bucket = my-bucket
+   path = /parsedmarc
+
 The full set of configuration options are:
 
 - ``general``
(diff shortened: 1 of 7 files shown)


docs/index.rst
```diff
@@ -200,7 +204,9 @@ The full set of configuration options are:
     - ``subject`` - str: The Subject header to use in the email (Default: parsedmarc report)
     - ``attachment`` - str: The ZIP attachment filenames
     - ``message`` - str: The email message (Default: Please see the attached parsedmarc report.)
-
+- ``s3``
+    - ``bucket`` - str: The S3 bucket name
+    - ``path`` - int: The path to upload reports to (Default: /)
```
williamdes, line 209: Not an int

PR #225, opened 2021-02-12
Feature/custom/output/files

m0rcq:
This PR adds support for custom output filenames for CSV/JSON aggregates and forensics as:

--output-json-aggregate
--output-csv-aggregate
--output-json-forensic
--output-csv-forensic

While we can change the output directory via the -o flag this adds additional flexibility in use cases as:

* custom post-processing based on filename including metadata encoded in the filename
* use same output directory for different processing pipelines requiring different post-processing rules


m0rcq:
@seanthegeek: hope this PR's features will be helpful for the community if their use cases match the feature's intent.

Issue #238, opened 2021-04-19
Allow for limiting the number of messages processed via imap in a given run


drawks:
When pulling reports over imap as a batch job I'm finding that memory usage is unacceptably high and run times are unpredictably long with a large number of Inbox items. It would be great if there was a flag/option to limit the number of reports that would be pulled over imap in a given run.

mback2k:
I would also appreciate this feature being implemented.

tom-henderson:
I've also submitted #222 a while ago to achieve this.

PR #239, opened 2021-04-19
Resolve issue #238 - Limit number of imap reports

drawks:
* adds `limit_messages` configuration parameter to imap
* updates `README.rst` to reflect new parameter

It would be IMHO better if the code was restructured in such a way as to make this a global option and/or for the saving/exporting and the reading/processing parts to be run concurrently instead of serially.  That said, this does allow limiting the size of batch jobs to enable more predictable runtime and memory usage as when run as a periodic job over a report volume that may be highly variable.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -155,6 +155,7 @@ The full set of configuration options are:
     - ``watch`` - bool: Use the IMAP ``IDLE`` command to process messages as they arrive
     - ``delete`` - bool: Delete messages after processing them, instead of archiving them
     - ``test`` - bool: Do not move or delete messages
+    - ``limit_messages`` - int: Process a maximum of this many messages per run
 - ``elasticsearch``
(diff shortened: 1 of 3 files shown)


Kuzuto:
Hi, this pr would be very nice to get implemented.
I tested your changes on 6.12 version, but it looks like it can't do len(messages)
Getting this error when I stated Parsedmarc with Mailbox watch enabled:

    if not test and (len(messages) < limit_messages) and total_messages > 0:
TypeError: '<' not supported between instances of 'int' and 'NoneType'

seanthegeek:
@drawks  Sorry I've been very neglectful of these PRs. I just merged a bunch of them. I think this issue was already resolved by PR #222 ?

seanthegeek:
@drawks  I'm finally getting around to updating this project. Is this PR outdated by the `batch_size` option?

PR #245, opened 2021-06-06
Use UTC datetimes for Elastic output

mwander:
Elastic expects UTC datetimes. However, parsedmarc outputs local time. As consequence, Elastic will shift the date range erroneously by the local timezone offset.

The following example ...
  <date_range>
    <begin>1622851200</begin>
    <end>1622937599</end>
  </date_range>

... will be imported into Elastic as ...
    "date_range": [
      "2021-06-05T02:00:00.000Z",
      "2021-06-06T01:59:59.000Z"
    ],

... while it should be:
    "date_range": [
      "2021-06-05T00:00:00.000Z",
      "2021-06-05T23:59:59.000Z"
    ],

Note the "Z" at the end, which indicates that Elastic is using UTC timezone.

This pull request converts the dates in Elastic output to UTC. We add "Z" explicitly, though it is not necessary with Elastic's default, but it increases robustness in case when the default changes or is reconfigured.

### DIFF
--- a/parsedmarc/elastic.py
+++ b/parsedmarc/elastic.py
@@ -317,8 +317,8 @@ def save_aggregate_report_to_elasticsearch(aggregate_report,
     org_name_query = Q(dict(match_phrase=dict(org_name=org_name)))
     report_id_query = Q(dict(match_phrase=dict(report_id=report_id)))
     domain_query = Q(dict(match_phrase={"published_policy.domain": domain}))
-    begin_date_query = Q(dict(match=dict(date_range=begin_date)))
-    end_date_query = Q(dict(match=dict(date_range=end_date)))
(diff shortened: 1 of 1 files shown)


mwander:
I just realized, `begin_date_human` and `end_date_human` is used as error message only and never fed into Elastic. Thus, the added "Z" is superfluous (though won't hurt).

mwander:
Commit  4bc7b0b fixes the deduplication feature for Elastic output. Duplicate reports will be found and ignored, if the following fields match:
* org_name
* report_id
* published.policy_domain
* begin_date
* end_date

As an example for the need, Google regularly sends duplicate reports with identical report-id (and identical message-id, for what it's worth).

PR #252, opened 2021-06-19
fix what was broken in merge train

knightian:
Seems that a comma was missed, and an argument was included twice in the last merge train.

I fixed it :D


seanthegeek:
Thank you!

Issue #111, opened 2019-09-23
Microsoft stops with static passwords


hvdkooij:
Actually I am not sure if this needs to be a feature request.

but it seems Microsoft will stop supporting static passwords.
See also: https://techcommunity.microsoft.com/t5/Exchange-Team-Blog/Improving-Security-Together/ba-p/805892

What will be the impact on tools like parseddmarc?

seanthegeek:
I wasn't aware of this. Thanks for the heads up!

Yes, this will impact parsedmarc, because they are disabling Basic Authentication for IMAP

>Today, we are announcing we are also turning off Basic Authentication in Exchange Online for Exchange ActiveSync (EAS), POP, IMAP and Remote PowerShell at the same time – October 13, 2020.

It looks like they don't have a OAuth solution for IMAP clients yet

> So let’s talk about POP and IMAP. We know there’s still some usage out there, not much, but some. We’re planning on adding OAuth support to both POP and IMAP in the next few months.

I'll keep an eye out for more details from Microsoft later.

dinigalab:
Hi,

How can I configure Davmail for parsedmarc in order to use modern authentication?
I have seen the configuration you shared but i am not able to set the user and password.
here is my error message :
 ERROR:cli.py:606:IMAP Error: command: LOGIN => socket error: EOF

thanks

fraserc182:
Hi,

I was wondering if there is an update on this issue? Has MS got any updated solutions for IMAP clients?
Cheers.

dinigalab:
Hi,

I think MS support OAuth ( modern authentication).
But it was parsedmarc that only support IMAP.
That is why I opened the issue to the developer of parsedmarc tool.

Thanks

Téléchargez Outlook pour iOS<https://aka.ms/o0ukef>
________________________________
De : Fraser Clark <notifications@github.com>
Envoyé : Wednesday, April 15, 2020 8:58:28 AM
À : domainaware/parsedmarc <parsedmarc@noreply.github.com>
Cc : dinigalab <dini_galab@hotmail.com>; Comment <comment@noreply.github.com>
Objet : Re: [domainaware/parsedmarc] Add OAuth support for IMAP on Office 365 (#111)


Hi,

I was wondering if there is an update on this issue? Has MS got any updated solutions for IMAP clients?
Cheers.

noraisk:
I think it should be possible to start working with adding oauth support using specific python library:

[...]

PR #301, opened 2022-04-04
Add support for Microsoft Graph API (Microsoft 365 mailboxes)

nathanthorpe:
Add support for fetching messages from Microsoft 365 mailboxes using the Graph API (issue #111).

Note that this requires an app registration with admin consent to the `Mail.ReadWrite` (delegated) permission.

Generalize the `get_dmarc_reports_from_inbox` and `watch_mailbox` to support different types of mailboxes with the `MailboxConnection` interface. This could be used for the getting messages from the Gmail API without having to write a separate get reports method.


kohtala:
Hmm. This repository does not seem to use tags for versions, releases. Commit messages do not carry much useful info. I searched if semantic versioning is mentioned and looked at the CHANGELOG.md, but I did not see anything that would tell major and minor version numbers work by the semantic versioning rules. If semantic versioning is used, the changes as they are would require major number increase since it may break old installations unless configuration is modified.

I'm here because I commented on the issue about a change towards this, but I have not used parsedmarc yet. I use Google, so I can not take this into use with this alone. Thanks anyway, seems a step forward for me too.

Otherwise quickly skimming, I do not see other problems but the breaking change.

noraisk:
I was able to build the PR, but I wasn't able to run it.

```
# ./virtual/environment/bin/parsedmarc -c parsedmarc.ini
   ERROR:cli.py:598:You must supply input files, or an IMAP configuration
```

cat parsedmarc.ini
```
[general]
save_aggregate = True
save_forensic = True
offline = True
debug = True
n_procs = 1
chunk_size = 1

[elasticsearch]
hosts = https://<secret1>:9200/
ssl = True
cert_path = /etc/ssl/certs/ca-certificates.crt
monthly_indexes = True

[mailbox]
test = true

[msgraph]
client_id = <secret2>
client_secret = <secret3>
mailbox = <secret@example.com>

```

nathanthorpe:
> @nathanthorp can you refer what is missing here or share your working template of your configuration file?

Hmm I have changed the error message on this line L639 of cli.py to `You must supply input files, or a mailbox connection`. From looking at your output, it seems to still have the older version of the code.

How did you build it on your end?

Also you will need to give it a username and password since it uses delegated permissions instead of application permissions. I *could* implement it using application permissions but then you'd need to grant it read/write access to all mailboxes, which is less secure.

seanthegeek:
@nathanthorpe It looks like you used some syntax that requires Python >=3.8. Can you fix it to work with Python 3.6, so it will work with older Linux distributions?

https://github.com/domainaware/parsedmarc/runs/6119612789?check_suite_focus=true#step:7:27

nathanthorpe:
Done, also let me know if more documentation is required for this. I only updated the main README but I noticed there was also one in docs/

nathanthorpe:
I see that there was a PR related to Gmail, if you want I can try to merge it with the MailboxConnection interface to clean it up a little

seanthegeek:
Sure. If you could also run the code through flake8 before submitting the next PR, that would be great. Thanks!

seanthegeek:
> Also let me know if more documentation is required for this. I only updated the main `README` but I noticed there was also one in `docs/`

For the config listing in the index in docs, I just copy/paste from the `README`.

seanthegeek:
Actually, I just noticed that all of the pep8 errors are from the Gmail PR. I'll fix those.

PR #307, opened 2022-04-21
Refactor Gmail integration with MailboxConnection interface

nathanthorpe:
Changes:

- Use  MailboxConnection interface for Gmail related code to reuse existing `get_dmarc_reports` logic.
- README and param updates
- Remove deprecated `run_console` method when getting gmail creds, replace with `run_local_server` (https://developers.googleblog.com/2022/02/making-oauth-flows-safer.html?m=1#disallowed-oob")
- Other PEP8 fixes


seanthegeek:
The tests hit an error https://github.com/domainaware/parsedmarc/runs/6121079486?check_suite_focus=true

nathanthorpe:
Yeah still working on this PR, it will be draft until I get a chance to test it with Gmail.  I will keep an eye on the CI checks.

nathanthorpe:
Ok this is ready for review. Seems to work with gmail on my end.

exetico:
Hi

I finally took some time to setup parsedmarc and I should have read how it's working - cause now my 20+ mails in my Gmail categories is not visible in the categories anymore.

With that said;
I've added both imap and gmail_api as options, but would something point out why I need to enable both things? If I comment out the [imap] part, it's returning `ERROR:cli.py:598:You must supply input files, or an IMAP configuration` in the logs.

nathanthorpe:
> **Q: Why do I need to add both imap and gmail_api**?

You shouldn't have to supply both, it will only work with one or the other.

Looking at your log output it seems that you have an older version installed? Please verify that v8 is installed.

If you look at L663 of cli.py, the error message that you are referencing is: "You must supply input files or a mailbox connection"
https://github.com/domainaware/parsedmarc/blob/master/parsedmarc/cli.py#L662

exetico:
Thank you for the quick response. I ended up disabling parsedmarc on the machine, but I'll move it to another one. There was simply not enough room for more ram-usage on the instance. but I totally missed that fact that it will require a bit of both ram and CPU on/off 😄 ... And more than a few of my other running containers...

I was using the `patschi/parsedmarc:latest` tag.

nathanthorpe:
That's not an official image, so I'm assuming that it is an outdated version.

PR #320, opened 2022-05-12
Implement Device Code and Client Secret auth flow for MS Graph

nathanthorpe:
The `auth_method` and `tenant_id` properties are added to the `msgraph` config section. Defaults to *UsernamePassword*

**Device Code**

This authenticates as a user and supports accounts that have MFA enabled as it prompts the user to open up a browser. 

Example:
`To sign in, use a web browser to open the page https://microsoft.com/devicelogin and enter the code <CODE> to authenticate.`

**ClientCredential**

This authenticates as the app (service principal). Note that this needs application permissions (not delegated).
This is probably the best way to authenticate, if you restrict mailbox access.

**Other edits**
- Fix scope when using a shared mailbox when using the *UsernamePassword* auth method.
- Includes README updates for the Azure AD app configuration.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -221,14 +221,24 @@ The full set of configuration options are:
     - ``password`` - str: The IMAP password
 
 - ``msgraph``
-    - ``user`` - str: The M365 user
-    - ``password`` - str: The user password
+    - ``auth_method`` - str: Authentication method, valid types are UsernamePassword, DeviceCode, or ClientSecret (Default: UsernamePassword).
+    - ``user`` - str: The M365 user, required when the auth method is UsernamePassword
+    - ``password`` - str: The user password, required when the auth method is UsernamePassword
     - ``client_id`` - str: The app registration's client ID
     - ``client_secret`` - str: The app registration's secret
-    - ``mailbox`` - str: The mailbox name. This defaults to the user that is logged in, but could be a shared mailbox if the user has access to the mailbox
+    - ``tenant_id`` - str: The Azure AD tenant ID. This is required for all auth methods except UsernamePassword.
+    - ``mailbox`` - str: The mailbox name. This defaults to the current user if using the UsernamePassword auth method, but could be a shared mailbox if the user has access to the mailbox
 
     .. note::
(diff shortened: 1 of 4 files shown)


seanthegeek:
Funny, I actually had a branch that implemented this last weekend, but I deleted it, because it turns out using client/secret auth requires system API access (instead of being deligated to a user account), which would grant read/write of **all mailboxes* in an org, which is way to broad of a scope.

Instead, I'm thinking we could do browser-based auth? Have the user copy/baste the URL from SSH into a browser to do the SSO MFA login for that account.

https://github.com/Azure/azure-sdk-for-python/blob/main/sdk/identity/azure-identity/azure/identity/_credentials/browser.py

I just haven't had time to look into it more.

nathanthorpe:
Yeah that's what I thought too and why I didn't implement it initially, but turns out you can limit access using the `New-ApplicationAccessPolicy` cmdlet in the Exchange module. 

The Device Code flow implements the browser based auth, which is a little better because the Interactive Browser credential requires a redirect URI to be provided (requiring something to listen on a port)

seanthegeek:
Interesting! Still, I don't trust every user to follow best practices. 😅

I wonder if we can verify the restricted scope, and refuse to use the creds until the scope in narrowed?

nathanthorpe:
Yeah I'm guessing there may be something we can do, like listing mailboxes and check if it is different from the provided mailbox. I will do some testing

seanthegeek:
Thanks for putting so much work into this project! It's so nice not to be my own while issues and requests pile up.

nathanthorpe:
Hey thanks for sharing this awesome project! Happy to help.

nathanthorpe:
Hmm... I'm not finding a good way to do this, if you wanted to list mailboxes you'd have to grant the application _even more_ permissions such as the `User.Read.All` scope.

I did add a message in the readme about properly setting this up. Maybe that would be sufficient?

seanthegeek:
Oof. In that case, I'll merge these changes, and then add more details to the warnings.

Issue #330, opened 2022-06-30
msgraph DeviceCode auth reports error AADSTS7000218 despite providing client_secret


jdghub:
I have upgraded to ParseDMARC 8.3.0 to use msgraph authentication instead of IMAP, but UsernamePassword doesn't seem to work (because the account has MFA enabled) and ClientSecret was getting complicated (because the account was created as a shared mailbox, though direct login is now enabled). So I would like to use DeviceCode authentication. I've:
- registered ParseDMARC as as an app
- recorded the Application (client) ID and the Directory (tenant) ID
- assigned the Mail.ReadWrite delegated permission
- created and recorded a Client Secret for the registered app

My INI file has
`[msgraph]`
`auth_method = DeviceCode`
`tenant_id = <tenant-id>`
`client_id = <client-id>`
`client_secret = <client-secret-value>`
`mailbox = <mbox>@<domain>`

When I run ParseDMARC I'm prompted to login with a code, which I do as `<user>@<domain>`, including MFA, and I accept to access ParseDMARC. But when I close the login window ParseDMARC exits with a long error stream that starts
`DeviceCodeCredential.get_token failed: Authentication failed: AADSTS7000218: The request body must contain the following parameter: 'client_assertion' or 'client_secret'.`
`Trace ID: b0378a2c-560e-44a2-8420-141397fcbe00`
`Correlation ID: e1c56f5d-c66d-48d8-8b47-717d151b205b`
`Timestamp: 2022-06-30 05:52:13Z`
`Content: {"error":"invalid_client","error_description":"AADSTS7000218: The request body must contain the following parameter: 'client_assertion' or 'client_secret'.\r\nTrace ID: b0378a2c-560e-44a2-8420-141397fcbe00\r\nCorrelation ID: e1c56f5d-c66d-48d8-8b47-717d151b205b\r\nTimestamp: 2022-06-30 05:52:13Z","error_codes":[7000218],"timestamp":"2022-06-30 05:52:13Z","trace_id":"b0378a2c-560e-44a2-8420-141397fcbe00","correlation_id":"e1c56f5d-c66d-48d8-8b47-717d151b205b","error_uri":"https://login.microsoftonline.com/error?code=7000218"}`
and then repeats 2 variants of the same info.

My INI file provides the client_secret value, and ParseDMARC errors out if it's not there. So is ParseDMARC not providing it for some reason, or is there something else I might have missed?

Thanks.

nathanthorpe:
[...]

PR #334, opened 2022-07-21
Add Graph credentials cache and fix issue with batch size

### LINKED ISSUE #333: Graph mail connector only pulls 10 messages per run
When running parsedmarc using the MSGraph mailbox connector, only 10 messages are processed per execution. I tried setting mailbox["batch_size"] to 50, but still only 10 messages were processed per execution. For reference I am running the latest version of parsedmarc available on pip, not running as a service. Maybe I this was a bad assumption, but I kind of thought that parsedmarc would process every message in the mailbox I pointed it to.

Per Microsoft documentation, Graph paginates return data from the .../messages endpoint, with default page size of 10 (https://docs.microsoft.com/en-us/graph/api/user-list-messages?view=graph-rest-1.0&tabs=http). So since there is currently no logic to handle pagination in parsedmarc/mail/graph.py in MSGraphConnection.fetch_messages, this is expected behavior on the MS side at least. 

I think there's a few options here, eg have fetch_messages return a generator of result pages (although this might mess up other mailbox connection types,) or have fetch_messages run through all the pages and return a full list of messages, or let the configured batch_size determine the page size returned from graph by adding a top=batch_size parameter to the URL used in fetch_messages.

If I can find a good way to test this out I would try doing a pull request, I'm not super GitHub literate though and I've already just run parsedmarc multiple times in my own environment to get through our messages :).

nathanthorpe:
Should resolve #333 and #330

- Add the ability to cache credentials from Graph (useful for the `DeviceCode` flow where you don't want to log in each time). This could probably deprecate the `UsernamePassword` flow since the other two are more ideal and work with MFA
- Add the `batch_size` arg into `fetch_messages` method since Graph defaults to only returning 10 messages
- Fix access denied error if using `DeviceCode` flow and connecting to a shared mailbox with a different account.
- Fix gmail configuration properties on readme and package requirement

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -228,6 +228,7 @@ The full set of configuration options are:
     - ``client_secret`` - str: The app registration's secret
     - ``tenant_id`` - str: The Azure AD tenant ID. This is required for all auth methods except UsernamePassword.
     - ``mailbox`` - str: The mailbox name. This defaults to the current user if using the UsernamePassword auth method, but could be a shared mailbox if the user has access to the mailbox
+    - ``token_file`` - str: Path to save the token file (Default: .token)
 
     .. note::
         You must create an app registration in Azure AD and have an admin grant the Microsoft Graph ``Mail.ReadWrite`` (delegated) permission to the app.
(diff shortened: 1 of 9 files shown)


seanthegeek (commented):
Thanks!

noraisk:
For me batch_size = 0 under [mailbox] neither [msgrpah] block didn't work. 
Maximum I've got working was 1000 and under [mailbox].
I shared my working example on 8.3.2 version here: https://github.com/domainaware/parsedmarc/issues/365#issuecomment-1293363572

PR #349, opened 2022-09-08
Handle unexpected xml parsing errors more gracefully

### LINKED ISSUE #348: Uncaught exception if xml report is an empty string
In this block https://github.com/domainaware/parsedmarc/blob/master/parsedmarc/__init__.py#L225 if `xml` is an empty string the parse operation will fail and remain uncaught, it will result in the cli.py raising/logging an error and exiting with a non-zero return code like so:

```
   ERROR:cli.py:618:IMAP Error: Document is empty, line 1, column 1 (<string>, line 1)
```

The parsing operation should guard against this by setting xml to a valid "zero value" on any parsing errors, something like `'<a/>'`.

drawks:
* updates `parse_aggregate_report_xml` to not raise an unhandled exception on parsing errors
* Resolves #348

### DIFF
--- a/parsedmarc/__init__.py
+++ b/parsedmarc/__init__.py
@@ -222,11 +222,14 @@ def parse_aggregate_report_xml(xml, ip_db_path=None, offline=False,
         xmltodict.parse(xml)["feedback"]
     except Exception as e:
         errors.append("Invalid XML: {0}".format(e.__str__()))
-        tree = etree.parse(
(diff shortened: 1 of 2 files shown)


drawks:
Took me a few tries to get through unit tests and codecoverage report. IMHO the tests for this project should be broken down into smaller units. At any rate, this change should catch where the report is empty and handle it as an invalid report instead of raising an uncaught exception.

Issue #356, opened 2022-09-26
Dashboard date_histogram error with elasticsearch 8.0 or higher


ykcab:
When using Elastic search version `=>8.0`, the is an error for the date interval
`Elasticsearch err:[1:512] [date_histogram] unknown field [interval] did you mean [fixed_interval]?`

Is there an updated query (json) file? or any configuration I should fix either in grafana or elasticsearch?
<img width="833" alt="Screen Shot 2022-09-26 at 1 12 32 PM" src="https://user-images.githubusercontent.com/16911521/192349995-50708e61-ff07-4249-a1ab-333225378f61.png">

TafkaMax:
Same issue.

ykcab:
Good news, I believe to have fixed this issue. I am testing this out with Elastic 8.3.4 & 8.4 and Grafana 9.0

msizec:
hi
I found out that some panels won't work anymore then, sayaing the interval must be non null non empty
Some panels are OK, some are not.

andryushchenko:
![image](https://user-images.githubusercontent.com/47642811/209920355-1fc2bda5-9fde-4830-99e4-57378a3d78c9.png)
Same here

dijitali:
A workaround has been outlined in https://github.com/debricked/dmarc-visualizer/issues/41#issuecomment-1425415494:

```dockerfile
RUN sed -i 's/fixed_interval/interval/g' /var/lib/grafana/dashboards/Grafana-DMARC_Reports.json
```

Looks like this was partly fixed with #366 but there are a few remaining places needing changing to support [elasticsearch v8 breaking changes](https://www.elastic.co/guide/en/elasticsearch/reference/master/migrating-8.0.html):

> The `date_histogram` aggregation’s interval parameter is no longer valid.
> 
> **Details**
> It is now an error to specify the interval parameter to the [`date_histogram` aggregation](https://www.elastic.co/guide/en/elasticsearch/reference/master/search-aggregations-bucket-datehistogram-aggregation.html) or the [composite `date_histogram` source](https://www.elastic.co/guide/en/elasticsearch/reference/master/search-aggregations-bucket-composite-aggregation.html#_date_histogram). Instead, please use either `calendar_interval` or `fixed_interval` as appropriate.

seanthegeek:
[...]

PR #366, opened 2022-10-17
Updated Grafana query template to support Elasticsearch 8.0 and higher

ykcab:
This template update adds support for Elasticsearch `8.3.x` and above.

When using elasticsearch `^8.0`, the error as stated in issue #356 is resolved by replacing `interval` with `fixed_interval`

Reference https://github.com/grafana/grafana/issues/19911

### DIFF
--- a/grafana/Grafana-DMARC_Reports.json
+++ b/grafana/Grafana-DMARC_Reports.json
@@ -156,7 +156,7 @@
         "y": 2
       },
       "id": 6,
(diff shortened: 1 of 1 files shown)


seanthegeek:
Thanks!

casperdcl:
cross-ref: https://discuss.elastic.co/t/invalid-interval-specified-must-be-non-null-and-non-empty/283878

PR #398, opened 2023-04-21
Fixed Bug in graph.py

rubeste:
Fixed Bug regarding the finding of a specific folder. This Bug caused parsedmarc to crash if it could not find the folder in one Ms Graph request. This is only an issue if your MailBox contains 10+ folders. It was solved by adding the `$filter=displayName eq '{folder_name}'` param so it would immediatly find the folder.


rubeste:
Fixed Bug regarding the obtaining of messages from a folder. This Bug caused parsedmarc to only load the first 10 e-mails if the `batch_size` is defined as `0`. Now it will loop until the json parameter `@odata.nextLink` is not present.

PR #406, opened 2023-05-07
Change publicsuffix2 to publicsuffixlist

jlaundry:
The public_suffix_list.dat included in the `publicsuffix2` package is now quite out of date.

While one _could_ use `get_base_domain()` with `use_fresh_psl=True`, this isn't done by `parse_aggregate_report_xml()`, and would be a big performance hit to download the PSL each time.

The `publicsuffixlist` package now includes a weekly updated psl.dat (https://github.com/ko-zu/psl/issues/25), **and** handles reserved domains sensibly out-of-the-box.


seanthegeek:
Looks like some unused imports need cleaned up.

jlaundry:
Oops, should be fixed now

Issue #282, opened 2022-01-10
Ignore RFC-incompliant date ranges


RealSebFox:
We receive DMARC reports from Trustwave SEG with date ranges exceeding 24 hours. Sometimes they even cover multiple weeks, as in this example:

```
<date_range>
   <begin>1637392339</begin>
   <end>1639516233</end>
</date_range>
```

Date ranges excessively exceeding 24-48 hours make data look distorted in Kibana. Since https://datatracker.ietf.org/doc/html/rfc7489#section-7.2 clearly mentions that _Visibility comes in the form of daily (or more frequent) Mail_, it would probably be best to consider such reports "invalid".

RealSebFox:
![Screenshot from 2022-03-01 23-00-22](https://user-images.githubusercontent.com/47162552/156256406-70bc64a8-f863-4bef-90da-e3ae96c9a90f.png)

Here is an example of the mentioned distortion. I'm going to trace them back manually by exclusion and add filters to exclude the culprits for good. That is a tedious job. Such RFC-incompliant reports keep popping up.

RealSebFox:
Hello @seanthegeek,

Yesterday, I upgraded from 8.6.1 to 8.6.4 and the problem has returned.

It was broken with this change: https://github.com/domainaware/parsedmarc/blob/master/CHANGELOG.md#863.

Please make sure that invalid or broken reports are not sent to elasticsearch, because they are not RFC compliant and it distorts the timelines!

Thank you!

PHPGangsta:
Hello,

For me the problem is also back, graphs in Kibana are broken again after 8.6.3 because invalid reports get imported to ElasticSearch again.

https://github.com/domainaware/parsedmarc/commit/1655b84cc28202de2f6989939269ae182a0ebfdf#diff-9234afbce8b159da96ee1a50bdca3307fc6f2cfbb7ac00f1fd41120c5a0d1a10R278

@seanthegeek Can you please explain why the Exception has been changed to only an error? Is it possible to have an error and DON'T import the report to ElasticSearch? I don't care if it's an error or an exception, important is that invalid reports don't get imported into ElasticSearch.

Here a screenshot with broken graphs in Kibana. I selected "Last 1 month", but because some invalid reports did contain data of multiple month in a single report, it does not only show 1 month, but a graph starting in February:
![grafik](https://github.com/user-attachments/assets/63add7f8-468e-4aeb-b3db-fd20c912a4a7)

[...]

PR #408, opened 2023-05-09
Fix issue #282: Detect large date ranges in aggregate reports and ski…

PHPGangsta:
…p processing


seanthegeek:
Thanks to you both

PR #417, opened 2023-05-17
Fix: Less error-prone code to minimize the risk of program crash

andersnauman:
- Double check if org_name exist. Empty name will crash Elastissearch's 'exist' search
- Move try-statement to include open() to catch if files do not exist
- Enclose Elasticsearch's execute in a try-statement to catch any invalid searches when variables are empty/missing

### DIFF
--- a/parsedmarc/__init__.py
+++ b/parsedmarc/__init__.py
@@ -254,6 +254,13 @@ def parse_aggregate_report_xml(xml, ip_db_path=None, offline=False,
             new_org_name = get_base_domain(org_name)
             if new_org_name is not None:
                 org_name = new_org_name
+        if not org_name:
+            logger.debug("Could not parse org_name from XML.\r\n{0}".format(
+                report.__str__()
+            ))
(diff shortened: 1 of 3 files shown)


qwell:
I'm not sure if it's the same issue that prompted you to make this PR, but I was just having some problems with reports from Yahoo, where `org_name` is Yahoo, rather than yahoo.com or yahooinc.com.  When `get_base_domain()` gets called, it returns `None`.  The fallback of using the domain from `email` happens before checking whether `org_name` is `None`, which means a fallback never happens for an "invalid" `org_name`.

seanthegeek:
@andersnauman For some reason this is failing unit testing. I haven't looked into it yet.

andersnauman:
@seanthegeek You are correct :), although, I think those tests needs some overhaul since they have been lurking for failure for some time now. `def testAggregateSamples(self):` reads test-files which raises `InvalidDMARCReport`, but those exception are never caught in the test.

`
`

Issue #429, opened 2023-09-01
Invalid reports from mimecastreport.com due to trailing \r\n on gzip data


0xabu:
I've recently received some reports from Mimecast that parsedmarc flags as invalid. The root cause appears to be that they have an attachment, where the valid gzipped report data is followed by a spurious CRLF. This causes Python's GzipFile to reject them here:
```
File .../python3.9/site-packages/parsedmarc/__init__.py:392, in extract_xml(input_)
    391 elif header.startswith(MAGIC_GZIP):
--> 392     xml = GzipFile(fileobj=file_object).read().decode(errors='ignore')
    393 elif header.startswith(MAGIC_XML):

File .../lib/python3.9/gzip.py:300, in GzipFile.read(self, size)
    299     raise OSError(errno.EBADF, "read() on write-only GzipFile object")
--> 300 return self._buffer.read(size)

File .../lib/python3.9/gzip.py:487, in _GzipReader.read(self, size)
    486 self._init_read()
--> 487 if not self._read_gzip_header():
    488     self._size = self._pos

File .../lib/python3.9/gzip.py:435, in _GzipReader._read_gzip_header(self)
    434 if magic != b'\037\213':
--> 435     raise BadGzipFile('Not a gzipped file (%r)' % magic)
    437 (method, flag,
    438  self._last_mtime) = struct.unpack("<BBIxx", self._read_exact(8))

BadGzipFile: Not a gzipped file (b'\r\n')
```

However, the command-line `gzip` tools extract them without issue, and presumably other parsers handle these too.

I found some discussion of the limitations of `GzipFile` [here](https://stackoverflow.com/questions/4928560/how-can-i-work-with-gzip-files-which-contain-extra-data) -- maybe one of these workarounds would make sense for parsedmarc?

These reports came from:
```
<org_name>Mimecast</org_name>
<email>no-reply@au-1.mimecastreport.com</email>
<extra_contact_info>https://community.mimecast.com/s/knowledge</extra_contact_info>
```

Kuzuto:
Made a pull request #430 , that should fix the problem.

nhairs:
It may be worth adding tests for this if you're willing to provide a sample email @0xabu or by generating our own spurious sample.

0xabu:
Sure: [mimecast-bad-dmarc.eml](https://github.com/domainaware/parsedmarc/files/13805223/mimecast-bad-dmarc.eml.txt)

PR #456, opened 2024-01-02
Add additional samples and ensure that git does not modify samples

nhairs:
This PR:

- adds the sample from #429 to prevent regressions.
- adds a copy of `dmarc_ruf_report_linkedin.eml` with line endings converted to CRLF to ensure there is at least one forensic sample with the line endings.
- adds a `.gitattributes` file that ensures that the `samples/` directory is treated as binary data. Without this git would treat this data as text data and automatically convert between `/r/n` (CRLF / DOS) line endings and `/n` (UNIX) line endings when files are checked out.
  This is important for ensuring that our parsing logic can be tested on both types of line endings.

### Mimecast Regression
1. Checkout `1de4a94c3787d2470e4dacd8e0de1b4027bc0f21` (parent of commit that fixed mimecast issue)
2. Run tests and check that it fails with
  `parsedmarc.InvalidAggregateReport: Invalid archive file: Not a gzipped file (b'\r\n')`
3. checkout HEAD and run tests again, this time with no errors.

### Forensic sample with CRLF
1. Run `file samples/forensic/dmarc_ruf_report_linkedin.crlf.eml`
2. Check that the output says that it has "CRLF line terminators"

### Aggregate sample with CRLF
1. Run `file samples/aggregate/mimecast-weird-gzip.eml`
2. Check that the output says that it has "CRLF line terminators"

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,1 @@
+samples/* binary
--- a/samples/aggregate/mimecast-weird-gzip.eml
+++ b/samples/aggregate/mimecast-weird-gzip.eml
@@ -1,0 +1,132 @@
+Received: from CY8PR05MB9817.namprd05.prod.outlook.com (::1) by
+ BL0PR05MB4724.namprd05.prod.outlook.com with HTTPS; Thu, 31 Aug 2023 10:06:17
+ +0000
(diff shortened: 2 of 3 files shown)


seanthegeek:
Thanks!

Issue #455, opened 2024-01-02
Add Type Hints


nhairs:
It would be good if `parsedmarc` added and exposed typing information so that calling libraries can check they are passing the correct types.

In theory this shouldn't be too difficult as the doc-strings of most functions contain type information.

To support inline type hints would require dropping support for `python<3.5`, however given the [security support schedule](https://endoflife.date/python) it may be desirable to drop support up to `3.7`. [Checking PyPI Stats](https://pypistats.org/packages/parsedmarc), versions `<=3.6` generally makeup less than 1% of daily downloads (though there appears to be the odd spike to ~5-8% for python 3.6).

![image](https://github.com/domainaware/parsedmarc/assets/9976336/236bdf20-9719-46c0-a731-451eb64f10f3)

In addition we'd want to add a `requires-python>=3.7` to `pyproject.toml` to prevent older versions attempting to install a version they would be unable to run.

I'd be willing to help out with this.

nhairs:
I've begun working on this at https://github.com/nhairs/parsedmarc/tree/typing

I'll note that there is more work than just adding type hints in that branch. At the same time I've done some refactoring, and applied the `black` code formatter (with line length 100). The end result should be more readable code that is still functionally the same as it was before adding any type hints.

PR #458, opened 2024-01-09
Add typing / drop Python < 3.7 support

nhairs:
This PR adds Python 3.7+ compatible type annotations (dropping support for <3.7 at the same time).

fixes: #455 

At a high level the following changes have been made:
- Type annotations
- Use newer python features / standard library where available
- Run the `black` and `isort` formatter

## Test Plan

1. Ensure that `flake8` without errors
2. Ensure that `python3 tests.py` passes without errors
3. Ensure that `mypy` passes without errors.

## Open Questions

Q1: Currently I've left all instances of `OrderedDict` alone as I'm not sure if we're using other features of the class. If we're only using the fact that they keep insertion order we should convert them all to `dict` since as of 3.7 they are guaranteed to retain insertion order.

That said, I think it would be better to move to using something like Dataclasses or Pydantic to properly type the report objects.


pyproject.toml
```diff
@@ -28,6 +28,9 @@ classifiers = [
     "Operating System :: OS Independent",
     "Programming Language :: Python :: 3"
 ]
+
+requires-python = ">=3.7"
+
 dependencies = [
```
nhairs, line 34: TODO: Add type stubs to dev dependencies

nhairs:
This is mostly done, just need to cleanup ready for final PR.

CC: @seanthegeek

nhairs:
@seanthegeek - I believe that CI is failing because Elasticsearch 8 cannot be used without a password. Should we make CI:
- use ES7?
- configure ES8 with static password?

nhairs:
> ## [Codecov](https://app.codecov.io/gh/domainaware/parsedmarc/pull/458?src=pr&el=h1&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=domainaware) Report
> Attention: `161 lines` in your changes are missing coverage. Please review.
> ....

Looks like this is mostly because there are overall more lines of code so even though it functionally has barely changed, the percentage of lines covered has actually dropped.

I think we can ignore the coverage change in this instance.

seanthegeek:
I'm not willing to drop Python 3.6 support for a while because RHEL 8 and its derivatives use Python 3.6. The end of life of most popular RHEL 8 derivative, Rocky Linux 8 isn't until May 31, 2029. That said, I'd like to include as much typing information is possible while still maintaining support for RHEL 8 and its derivatives.

nhairs:
Given that [postponed annotations](https://peps.python.org/pep-0563/) (`from __future__ import annotations`) is only supported in 3.7+, I don't think this diff can easily be made compatible with 3.6 typing (nor am I willing to spend the effort). As a result I'm closing this PR.

It might be worth formalising the supported versions and including / linking to them in [the contributing docs](https://domainaware.github.io/parsedmarc/contributing.html).

As an aside: I've decided to continue working on my forked repository to experiment with a new way of running the application. Given that it will not be python 3.6 compatible I don't think it will be possible to merge it (in addition to the many breaking changes). Assuming my experiment is successful I will end completely forking it (including changing the name and releasing it on PyPI). If you wish to discuss this, feel free to reach out to me directly.

Issue #464, opened 2024-01-23
ModuleNotFoundError: No module named 'msgraph'


laughk:
## Description

I have installed parsedmarc but it does not work with the following error:

```console
$ parsedmarc --help
Traceback (most recent call last):
  File "/tmp/parsedmarc/bin/parsedmarc", line 5, in <module>
    from parsedmarc.cli import _main
  File "/tmp/parsedmarc/lib/python3.11/site-packages/parsedmarc/__init__.py", line 31, in <module>
    from parsedmarc.mail import MailboxConnection
  File "/tmp/parsedmarc/lib/python3.11/site-packages/parsedmarc/mail/__init__.py", line 2, in <module>
    from parsedmarc.mail.graph import MSGraphConnection
  File "/tmp/parsedmarc/lib/python3.11/site-packages/parsedmarc/mail/graph.py", line 10, in <module>
    from msgraph.core import GraphClient
ModuleNotFoundError: No module named 'msgraph'
```

I am seeing this error in at least the following environments

- ArchLinux ( `Linux 5.15.133.1-microsoft-standard-WSL2 #1 SMP Thu Oct 5 21:02:42 UTC 2023 x86_64 GNU/Linux` ), Python 3.11
- Ubuntu 22.04.3 （`Linux 6.2.0-1017-aws #17~22.04.1-Ubuntu SMP Fri Nov 17 21:19:35 UTC 2023 aarch64 aarch64 aarch64 GNU/Linux` ）, Python 3.10

## Steps to reproduce

This occurs when you install parsedmarc and run the command in the following flow.

```console
$ python -m venv venv
$ source venv/bin/activate
(venv) $ python -m pip install parsedmarc
```

Running `parsedmarc` in this state will result in the following

```console
$ parsedmarc
Traceback (most recent call last):
  File "/tmp/parsedmarc/bin/parsedmarc", line 5, in <module>
    from parsedmarc.cli import _main
  File "/tmp/parsedmarc/lib/python3.11/site-packages/parsedmarc/__init__.py", line 31, in <module>
    from parsedmarc.mail import MailboxConnection
  File "/tmp/parsedmarc/lib/python3.11/site-packages/parsedmarc/mail/__init__.py", line 2, in <module>
    from parsedmarc.mail.graph import MSGraphConnection
  File "/tmp/parsedmarc/lib/python3.11/site-packages/parsedmarc/mail/graph.py", line 10, in <module>
    from msgraph.core import GraphClient
[...]

PR #466, opened 2024-01-25
fix: msgraph-core version

laughk:
fix https://github.com/domainaware/parsedmarc/issues/464

A breaking change was made in msgraph-core 1.0.0.
Currently, parsedmarc is not yet compatible with this version, so do not install 1.0.0 or higher.


seanthegeek:
I really need to check m y email more often. I just came across this as I was finishing up work for the SMTP-TLS reports. Thanks!

PR #496, opened 2024-03-25
removing obsolete parameter from cli_parse call

### LINKED ISSUE #495: Fix the cli_parse call
It looks like that the `cli_parse` call broke during one of the PR merges.
```
TypeError: cli_parse() takes 7 positional arguments but 8 were given 
```

Szasza:
Closes #495


seanthegeek:
Thanks. The good news it I haven't actually released 8.9.0 yet because I forgot my new dev system doesn't have my pypi token configured in hatch yet.

Issue #500, opened 2024-03-30
Security concern with the recently introduced CSV download


Szasza:
In https://github.com/domainaware/parsedmarc/commit/18f7508a1f09db554bc8a31bb26acc79d5c37c3f#diff-953601e25cfe2a31ac1b2602fcab0be206d5c7bb860165d66679f455ead64fcfR335 a direct download from GitHub has been introduced.

Unfortunately the download happens before the code is having a chance to use the map file sitting with the code The only ways to avoid this from happening are:
- using offline mode, which is not desirable in all cases, or
- passing in a map as the function parameter, which is not possible as this parameter is not exposed to the CLI

This means that if a malicious actor gains access to the GitHub repo, a manipulated CSV file will automatically be downloaded for all `parsedmarc` installations.

It would be ideal to use the local file as a first option. Alternatively, it would be great to be able to turn off the automatic download without rendering the whole `parsedmarc` run offline, or maybe being able to provide the download URL as a config option, so the file can be refreshed in a more controlled manner.

@seanthegeek if you are ok with it, I am happy to rework the reverse DNS map usage slightly to address the above.

seanthegeek:
I'm thinking we can add three options to the config file:

- `always_use_local_reverse_dns_map` - Disables the download of the reverse DNS map
- `local_reverse_dns_map_path` - Overrides the default local file path to use for the reverse DNS map
- `reverse_dns_map_url` - Overrides the default download URL for the reverse DNS map

That would allow users to pick exactly how they would like `parsedmarc` to load the map.

I built the current behavior to encourage others to contribute to the map, and to avoid users needing to upgrade the `parsedmarc` package just to get an updated map, or riequire me to make new releases just for updated maps for anything other than correcting errors (like the last few releases 😅).

kitzler-walli:
My instance still tries to download the csv from github even though I set this in my ini file:
```
always_use_local_files = True
local_reverse_dns_map_path = /opt/parsedmarc/base_reverse_dns_map.csv
```

Szasza:
@kitzler-walli is it possible that you use a `parsedmarc` version earlier than `8.11.0`, or not using the correct config file when invoking `parsedmarc`? As per https://github.com/domainaware/parsedmarc/blob/master/parsedmarc/utils.py#L340-L341 if `always_use_local_files` is set.

PR #501, opened 2024-03-30
IP_ADDRESS_CACHE cache improvements

rodpayne:
- Only add entries to the cache if the lookup was successful.  (The ones that fail may work next time.)
- Retain IP_ADDRESS_CACHE entries for up to four hours.  (It takes almost a day to import a day's reports, so the lookups don't need to be extremely fresh.)

### DIFF
--- a/parsedmarc/__init__.py
+++ b/parsedmarc/__init__.py
@@ -48,7 +48,7 @@ MAGIC_GZIP = b"\x1F\x8B"
 MAGIC_XML = b"\x3c\x3f\x78\x6d\x6c\x20"
 MAGIC_JSON = b"\7b"
 
-IP_ADDRESS_CACHE = ExpiringDict(max_len=10000, max_age_seconds=1800)
+IP_ADDRESS_CACHE = ExpiringDict(max_len=10000, max_age_seconds=14400)
 REVERSE_DNS_MAP = dict()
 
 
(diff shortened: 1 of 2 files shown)


parsedmarc/__init__.py
```diff
@@ -48,7 +48,7 @@
 MAGIC_XML = b"\x3c\x3f\x78\x6d\x6c\x20"
 MAGIC_JSON = b"\7b"
 
-IP_ADDRESS_CACHE = ExpiringDict(max_len=10000, max_age_seconds=1800)
+IP_ADDRESS_CACHE = ExpiringDict(max_len=10000, max_age_seconds=14400)
```
Szasza, line 51: It may worth making this a config file setting, defaulting to the initial `1800`, as different use-cases may warrant different values.

rodpayne: Yes, making it a config file setting may be a good enhancement, but it may take me a while to figure out how to do it.  I can take a look at it after the holiday.

I don't think that there are currently many people locked into the 1,800 setting since it was not working at all until last week.

seanthegeek: If someone would like to contribute a PR to add that option, be aware that although `IP_ADDRESS_CACHE` is only referenced twice:

https://github.com/domainaware/parsedmarc/blob/master/parsedmarc/__init__.py#L103

https://github.com/domainaware/parsedmarc/blob/master/parsedmarc/__init__.py#L897

configuration items related to report processing must go through several layers of function calls, such as when I just added the new function calls to address the concerns raised in #500.

https://github.com/domainaware/parsedmarc/compare/041296b1f1ba1d0b7b46c2d34f46dd3a2c1ef78c...acef7bdd6ed571f6c049ad6e2f181c0b97bf2e6a

Issue #503, opened 2024-04-01
Centralize the configuration handling


Szasza:
As `parsedmarc`'s functionality gets richer, the functions' parameter lists are getting longer due to the config being parsed on top level and all the values are individually passed down in the call chain. This will yield enormous parameter lists over time.

It would be great to extract the configuration into a standalone class, so it could be included in any depth and accessed in a more lightweight manner, reducing the amount of passed parameters.

seanthegeek:
I got excited when I saw this subject thinking this was a PR. 😝

This is absolutely needed. It's a mess already! I'm just not sure what is the right way to redesign it. So many people have contributed code over the years that most of the code isn't mine anymore. It's a pain for me to get around the code, let alone someone brand new to the project.

seanthegeek:
I just reread your post. That design could work!

Szasza:
@seanthegeek I will work on the implementation then.

seanthegeek:
Thinking about it more, having all options as a config object would break things for people who are using specific functions of `parsedmarc` . How about a config dictionary instead that could be passed to existing functions as `**kwargs`?

Szasza:
@seanthegeek it can be done, however it will make the functions slightly more complex. I think it may be a better idea to think about which parts/functions of parsedmarc are used as standalone.

Is parsedmarc meant to be a self-contained app or a utility library?

seanthegeek:
Ideally it should be both, with some core functions used for parsing reports usable as a library, with some parts on top of that like mailbox monitoring being standalone. Separating those into `_` prefixed modules is going to take some time. I'll give it more thought when I have some time this weekend.

rodpayne:
With the configuration in a standalone class whose instance is passed to all functions, how about including the global dictionaries and caches (such as `IP_ADDRESS_CACHE`, `REVERSE_DNS_MAP`, etc.) so that every function can access them with only the one reference parameter being passed down the chain?  For example, if every function had an argument of `parsedmarc_globals`, the function could use `parsedmarc_globals.opts.dns_timeout`, `parsedmarc_globals.ip_address_cache`, etc., as needed)

Szasza:
@seanthegeek I think @rodpayne's proposal is close to what you would like to see.

lingfish:
Would something like dynaconf help here?

Szasza:
@lingfish I would think so. @seanthegeek thoughts?

seanthegeek:
[...]

PR #504, opened 2024-04-04
Use the syslog2 package to correctly send RFC3164 formatted syslog output

lingfish:
The format that #263 sends isn't technically strict RFC3164; it's more of a raw output, but because it uses Python's `SysLogHandler`, the `PRI` is sent, but the `HEADER` is not.

Merging this in will probably break existing use of this.

Ideally, `parsedmarc` should also have GELF output, possibly rename the existing feature from #263 to some more like "Raw", and implement this PR.


seanthegeek:
Yeah, so many people use this project for such a variety of different use cases that I'd rather not break existing functionality. Similar to why I want to be so careful about what I and others do to address #503.

Please update this PR to:

- Address the [code style issue](https://github.com/domainaware/parsedmarc/actions/runs/8549063837/job/23423771046)
- Add the `syslog2` to the [dependencies list](https://github.com/domainaware/parsedmarc/blob/master/pyproject.toml#L31) in `pyproject.toml`
- Add a `format` option under the `syslog` configuration group, with the current `raw` format being the implicit default, and `rfc3164` using syslog2
- Add details about the configuration options to the [documentation](https://github.com/domainaware/parsedmarc/blob/master/docs/source/usage.md), and set `format = rfc3164` in the example configuration file

lingfish:
Yeah, all fair points, I kinda mangled this one, and I don't think that `syslog2` dep is a good one.. the author has weirdly put debug statements throughout it using `print()`.  I shall work on it.

lingfish:
I'm aborting this PR in favour of implementing GELF logging instead.

PR #522, opened 2024-05-03
Fixes for kafka integration

ChibangLW:
Some variables were not set correctly.

### DIFF
--- a/parsedmarc/cli.py
+++ b/parsedmarc/cli.py
@@ -845,16 +845,8 @@ def _main():
                 exit(-1)
             if "user" in kafka_config:
                 opts.kafka_username = kafka_config["user"]
-            else:
-                logger.critical("user setting missing from the "
-                                "kafka config section")
-                exit(-1)
             if "password" in kafka_config:
                 opts.kafka_password = kafka_config["password"]
-            else:
-                logger.critical("password setting missing from the "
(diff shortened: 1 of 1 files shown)


CmellierLNDS:
I have this error while running kafka configuration : 
`CRITICAL:cli.py:870:forensic_topic setting missing from the splunk_hec config section`
I think it comes from this snippet at line 867. Maybe this code is there for a reason that i don't know but there is no mention on `smtp_tls_topic` in the documentation of kafka, can you enlighten me ?
Why do you need : 
```
            if "smtp_tls_topic" in kafka_config:
                opts.kafka_username = kafka_config["smtp_tls_topic"]
                opts.kafka_smtp_tls_topic = kafka_config["smtp_tls_topic"]
            else:
                logger.critical("forensic_topic setting missing from the "
                                "splunk_hec config section")
```

ChibangLW:
The text in the error message is wrong. The setting is used if you also enabled `save_smtp_sts` in the `general` section. 

If I have time I will add the config option to the documentation and only log the error if  `save_smtp_sts` is enabled (same for `save_aggregate` and `save_forensic`).

BTW your snipped is a bit outdated as `opts.kafka_username = kafka_config["smtp_tls_topic"]` was removed in this PR.

PR #532, opened 2024-06-28
Add GELF support

lingfish:
Implement the ability to log to a GELF server/input, via the use of [pygelf](https://github.com/keeprocking/pygelf).

This is my initial stab at it; similar to `SyslogClient`, I opted to use `parsed_aggregate_reports_to_csv_rows`.

This is all sent in one field to GELF (`parsedmarc`), where then it is normal practice to split the sub-fields out within your logging app (eg. Graylog).

This would be an example Graylog pipeline rule:

```
rule "DMARC GELF"
when
  has_field("parsedmarc")
then
  let json = parse_json(to_string($message.parsedmarc));
  let map = to_map(json);
  set_fields(map, "dmarc_");
end
```

### DIFF
--- a/docs/source/usage.md
+++ b/docs/source/usage.md
@@ -98,6 +98,11 @@ path = parsedmarc
 [syslog]
 server = localhost
 port = 514
+
+[gelf]
+host = logger
+port = 12201
+mode = tcp
 ```
 
 The full set of configuration options are:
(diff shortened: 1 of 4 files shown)


seanthegeek:
Thanks. Can you add documentation updates to this PR?

lingfish:
> Thanks. Can you add documentation updates to this PR?

Err, forgive me... I thought I had?

lingfish:
Hi @seanthegeek just wondering if we're all good here?

PR #550, opened 2024-08-30
Fix the tilde expansion for user home folder

### LINKED ISSUE #485: Tilde expansion in save path not honored
This could be just handled in documentation.
I'd set the ini output directory with a tilde which I expected to expand to my home directory
(I'm still working out how dmarc in general, and parsedmarc in particular, works)
Instead, a directory "~/the/rest/of/the/path" was created (which took me a while to find).
You might consider some sort of shell variable expansion
Or make the need for a full path in clear in the documentation

fboaventura:
This will enable the tilde expansion to map the output dir to the home user using `~/path`. (Fix #485)


seanthegeek:
This should probably be added for the input paths too, right?

seanthegeek:
Never mind. Input paths are handled by glob.

PR #552, opened 2024-09-02
fix: use namespace for separator

### LINKED ISSUE #551: IMAP: Character not allowed in mailbox name: '/'
Running Parsedmarc against a Dovecot IMAP server gives the following error message in the logs:

```
  File "/usr/lib/python3.11/site-packages/parsedmarc/cli.py", line 1385, in _main
    reports = get_dmarc_reports_from_mailbox(
              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.11/site-packages/parsedmarc/__init__.py", line 1431, in get_dmarc_reports_from_mailbox
    connection.create_folder(aggregate_reports_folder)
  File "/usr/lib/python3.11/site-packages/parsedmarc/mail/imap.py", line 30, in create_folder
    self._client.create_folder(folder_name)
  File "/usr/lib/python3.11/site-packages/mailsuite/imap.py", line 305, in create_folder
    imapclient.IMAPClient.create_folder(self, folder_path)
  File "/usr/lib/python3.11/site-packages/imapclient/imapclient.py", line 1031, in create_folder
    return self._command_and_check(
           ^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.11/site-packages/imapclient/imapclient.py", line 1753, in _command_and_check
    self._checkok(command, typ, data)
  File "/usr/lib/python3.11/site-packages/imapclient/imapclient.py", line 1759, in _checkok
    self._check_resp("OK", command, typ, data)
  File "/usr/lib/python3.11/site-packages/imapclient/imapclient.py", line 1636, in _check_resp
    raise exceptions.IMAPClientError(
imaplib.IMAP4.error: create failed: [CANNOT] Character not allowed in mailbox name: '/' (0.001 + 0.000 secs).
```

Dovecot use `.` instead of `/` as subfolder separator. Altering the sourcecode for `get_dmarc_reports_from_mailbox()` to use `.` fixes the problem for me. A config option for separator would be useful.

gaige:
Use the IMAP namespace separator based on the information retrieved from the IMAP command to retrieve namespaces, falling back to '/' if the command isn't available or the results don't include the private namespace.

Resolves #551

### DIFF
--- a/parsedmarc/__init__.py
+++ b/parsedmarc/__init__.py
@@ -1416,10 +1416,15 @@ def get_dmarc_reports_from_mailbox(connection: MailboxConnection,
     aggregate_report_msg_uids = []
     forensic_report_msg_uids = []
     smtp_tls_msg_uids = []
-    aggregate_reports_folder = "{0}/Aggregate".format(archive_folder)
-    forensic_reports_folder = "{0}/Forensic".format(archive_folder)
-    smtp_tls_reports_folder = "{0}/SMTP-TLS".format(archive_folder)
-    invalid_reports_folder = "{0}/Invalid".format(archive_folder)
+    folder_separator = connection.get_folder_separator()
+    aggregate_reports_folder = "{0}{1}Aggregate".format(archive_folder,
+                                                        folder_separator)
+    forensic_reports_folder = "{0}{1}Forensic".format(archive_folder,
+                                                      folder_separator)
+    smtp_tls_reports_folder = "{0}{1}SMTP-TLS".format(archive_folder,
+                                                      folder_separator)
+    invalid_reports_folder = "{0}{1}Invalid".format(archive_folder,
+                                                    folder_separator)
 
     if results:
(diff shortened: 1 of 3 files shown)


seanthegeek:
A couple of things: 

- I don't see `get_folder_separator()` defined anywhere when a do a quick search of the repo. Where is it defined at?
- Let's set the folder separator once instead of calling `get_folder_separator()` for each path.

gaige:
@seanthegeek Yeah, somehow I missed the other files, they're in the mail classes, they should be there now.  I've updated to get the separator once.

gaige:
I have verified that this works in my configuration with Dovecot. The `IMAPClient` `namespace` method is used for determining the separator uses the IMAP standard call for determining the separator, so I don't expect any problems with other servers, however I've put in fallbacks for both IMAP and the base class to use the `/` separator which was used originally. In the case of the IMAP class, `/` will be returned if the call doesn't contain the expected parameters.

Issue #82, opened 2019-06-01
Does parsedmarc support reading emails from Maildir directly, instead of relying on IMAP server?


SkullTech:

seanthegeek:
No, But I can take this as a feature request :)

SkullTech:
Awesome! Thanks.

mda1125:
If IMAP isn't turned on for our configuration, I wonder if that mailbox can be accessed using Google's API.  The API could get mail whereas IMAP is disabled for the account.

That would be nice feature request.

seanthegeek:
@mda1125 I created a separate issue to track the Gmail feature request #96

seanthegeek:
Maildir input support was added in PR #555 (released in v8.15.1) via the `[maildir]` config section (`maildir_path`, `maildir_create`), and has since received several hardening fixes for Docker deployments. See https://domainaware.github.io/parsedmarc/usage.html for configuration details. Closing as implemented.

---
🤖 Triaged and closed by [Claude Code](https://claude.com/claude-code) at the maintainer's request.

PR #555, opened 2024-09-04
Added input report maildir connector, issue #82

spaolo:
Even working directly with maildir files may not be the finest way to handle mail messages may be useful when working on large numbers.
Please consider including this mailbox.Maildir wrapper


spaolo:
commit rebased

PR #567, opened 2024-10-05
Create optional dependency group for build, fix codecov

jlaundry:
Kia ora Sean,

I'm using parsedmarc to parse DMARC RUAs, but not using the built-in CLI. As a result, I get dependabot warning of outdated packages, as `msgraph-core`, `elasticsearch`, and `elasticsearch-dsl` are held back.

~~**This is a breaking change**, as people who are used to `pip install parsedmarc` will need to `pip install parsedmarc[cli]` instead. I know that's not ideal - I'm keen for your thoughts, and not merging this is a valid option 🙂~~

This PR:

- Creates a **build** optional dependency group, and updates the GitHub CI workflow, so that tests always run with the current package dependencies (the packages in `requirements.txt` weren't always the same as the `pyproject.toml`)
- ~~Creates a **cli** optional dependency group, so that those like me who don't use the CLI, don't need to install `elasticsearch` / `boto3` / `opensearch-py`~~
- Fixes codecov (which [stopped working in April](https://app.codecov.io/github/domainaware/parsedmarc)), provided that the `CODECOV_TOKEN` Actions secret is set

Future work:

- Update the `msgraph-core` package - I was thinking of going for this next, but need to get a test environment running first
- ~~Create a **mail** optional group (`msgraph-core`, `imapclient`, `google-*`) - that's not as easy, and would require a refactor of how `MailboxConnection` is imported~~


seanthegeek:
I would consider this change for a younger project, but I generally don't like breaking changes at this stage where the code is mature and [so frequently downloaded](https://pypistats.org/packages/parsedmarc).

The optional build group and codecov changes sound great though.

jlaundry:
Fair enough, I thought that might be the case - updated now.

Issue #572, opened 2024-10-22
Change deprecated panel made by angular to latest version made by react


yjszk:
![image](https://github.com/user-attachments/assets/620f6617-5602-4a49-9981-9994c63cee5b)

official info:
https://grafana.com/docs/grafana/latest/developers/angular_deprecation/

file:
https://github.com/domainaware/parsedmarc/blob/master/grafana/Grafana-DMARC_Reports.json

seanthegeek:
@yjszk Thanks for the call out. I use Splunk, so it would be great if someone else could complete the Grafana migration steps documented in the link above and contribute the changes here in a Pull Request.

PR #578, opened 2024-11-13
Add Grafana Dashbroad

ericericsw:
update new version dashbroad

panel model change list:
grafana-piechart-panel -> pie chart
Graph(old) -> time series
worldmap panel -> geomap

some table panel has change , be like overview add ARC Column

The problem cannot be solved at the moment: Multiple DKIM information will cause table display errors

### DIFF
--- a/grafana/Grafana-DMARC_Reports.json-new_panel.json
+++ b/grafana/Grafana-DMARC_Reports.json-new_panel.json
@@ -1,0 +1,5901 @@
+{
+  "annotations": {
+    "list": [
+      {
+        "builtIn": 1,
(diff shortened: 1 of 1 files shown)


ericericsw:
I saw someone mentioned that grafana dashbroad needs to be updated.
at https://github.com/domainaware/parsedmarc/issues/572

Sorry, I'm not very good at using github yet
If there are any changes that need to be made, please let me know

Supplement, I use grafana 10.1.6 because there seems to be an exception with higher versions of Ad hoc filters

seanthegeek:
Thanks!

PR #590, opened 2025-01-29
Add support for Microsoft national clouds via Graph API base URL

kevingoad-arcfield:
Added new configuration property for Graph API base URL
Updated documentation to explain usage of the new property

https://learn.microsoft.com/en-us/graph/deployments


kevingoad-arcfield:
Can someone test this with an Azure Commercial tenant?  I can confirm this works w/ Azure Gov.

PR #617, opened 2025-10-08
updates process_reports to respect newly added prettify_json option

macmandr197:
I want to facilitate the parsing of JSON logs directly with Grafana Alloy + Loki. To do this, it would be beneficial to output the JSON logs as a single line to be then parsed within Grafana. Currently, because the report is "prettified" it treats each new line of the log, as a different log entry within grafana. Collapsing the log to a single line, will allow it to be parsed properly within Grafana.

### DIFF
--- a/parsedmarc/cli.py
+++ b/parsedmarc/cli.py
@@ -620,7 +620,7 @@ def _main():
         la_dcr_forensic_stream=None,
         la_dcr_smtp_tls_stream=None,
         gelf_host=None,
-        gelf_port=None,p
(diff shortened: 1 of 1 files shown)


macmandr197:
> Let's remove that option to avoid cluttering the CLI with an option that doesn't do anything, because the fact that the JSON is pettified by default is implied by the existence of `--no-prettyify-json`, and the sample output in the documentation. Then I'll merge this.

Sounds good. I have removed the option. Let me know what you think.

PR #629, opened 2025-11-30
9.0.0

seanthegeek:
This ChatGPT-assisted PR normalizes aggregate DMARC report volumes when a report timespan exceeds 27 hours to solve two competing problems.

## Problem 1: Some receivers only send multi-day reports

If you reject them (like we do currently), parsedmarc is useless for those deployments.

With this design:

- Their data is included.
- It’s mathematically consistent with per-day charts.
- You still surface that the sender is non-standard via the normalization metadata.

## Problem 2: Leaving them raw causes huge distortions (issue #282)

If a 3-day report of 300 messages is dropped onto one day:

- That day appears to have 300 messages
- Neighboring days have 0 → huge spike.

With normalization records are expanded into multiple records:

- Day 1: ~100
- Day 2: ~100
- Day 3: ~100

`normalized_timespan`: boolean
`interval_begin`: timestamp
`interval_end`: timestamp

Are added to each each record

No more pathological peaks solely because of reporting span.

You can still a weird reporter by:

- Filtering documents where `normalized_timespan=true`, or
- Aggregating on `original_span_seconds` to see which reporters are misbehaving.

## Feedback requested

I'd like to get some feedback and additional testing from the parsedmarc user community before I merge this and release it.


Szasza (changes requested):
Looks good, apart from a minor version discrepancy.

seanthegeek:
@AnaelMobilia I'm seeing mixed information about if it is possible to query Elasticsearch/Opensearch for a date range in a subdocument. Ideally we would do a migration that just adds this on all existing reports (that have a valid date range because invalid report date ranges were rejected):

`normalized_timespan`: `false`
interval_begin: Same as the `date_begin` of the report
interval_end: Same as the `date_end` of the report

Release a new version of the Kibana and Graphana aggregate dashboards that uses `interval_begin` of the record, instead of `date_begin` of the report, but I don't know if that is possible.

PR #667, opened 2026-03-07
Add postgresql storage backend

fsvm88:
Hi, I've been using the project for a while, and in my 15 years of IT career (systems/cloud engineer), I always struggled with ElasticSearch.
I feel it's clunky and heavy, I don't understand the architecture, and every attempt at learning it pushed me away from it (pretty much like Kafka). It takes ~5m to start a single-node cluster on fast HDDs (so I have to work it around by making the `parsedmarc` container 5m before attempting to connect), and I always have to search the right command-line incantations and env variables to make it actually work. I'm unable to perform basic or periodic maintenance because I just can't understand it (and the curl-interface doesn't really help).

I recently ran into [this article](https://www.tigerdata.com/blog/its-2026-just-use-postgres) and it strongly resonated, so I decided to give it a shot and add a Postgres backend to this project as an alternative to ElasticSearch. Disclaimer: it's 99% LLM.

Claude Opus 4.6 basically just oneshotted the python code at the first try. I built a test docker image and ran it for the past week with zero issues so far (I have not received forensic reports or SMTP TLS reports though).

The Grafana dashboard went through a couple iterations to refine the queries, and it also works well for now (again, no forensic reports or SMTP TLS reports to show).

Postgres is much easier and transparent to manage, it's SQL, doesn't need customized command-line, I can perform maintenance and understand what's happening, and it's also very efficient:

| POD | CONTAINER | CPU | RAM |
|----|-----|-----|----|
| parsedmarc-5b56455f78-9tl6m | elasticsearch | 17m | 1750Mi |
| parsedmarc-5b56455f78-9tl6m | parsedmarc | 1m | 146Mi |
| parsedmarc-5b56455f78-9tl6m | postgres | 1m | 62Mi |

I hope it's a useful addition to the project. Let me know if you need me to change or fix anything :)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -27,6 +27,12 @@ Several elements that became `langAttrString` in RFC 9990 (`extra_contact_info`,
 
 Backwards compatibility to RFC 7489 is maintained.
 
+#### PostgreSQL storage backend
+
+New optional PostgreSQL output backend as a lighter-weight alternative to Elasticsearch/OpenSearch, configured via a `[postgresql]` section (host/port/user/password/database or a libpq `connection_string`). Tables are created automatically on first run, and the schema captures the RFC 9990 aggregate fields (`np`, `testing`, `discovery_method`, `generator`, `xml_namespace`, and per-result `human_result`). A Grafana dashboard (`grafana/Grafana-DMARC_Reports-PostgreSQL.json`) is included. Aggregate and SMTP-TLS reports are de-duplicated via `ON CONFLICT`; failure reports via an arrival-date / From / To / Subject check mirroring the Elasticsearch backend.
+
+The backend is opt-in: install it with `pip install parsedmarc[postgresql]` (it pulls in `psycopg`). It is not a mandatory dependency because the prebuilt `psycopg` binary wheels are not available for every platform.
+
 #### Docker-secret support via `_FILE` env vars
 
 Any `PARSEDMARC_{SECTION}_{KEY}` environment variable can now also be supplied via a file by appending `_FILE` to its name (e.g. `PARSEDMARC_IMAP_PASSWORD_FILE=/run/secrets/imap_password`). The file's contents (with trailing CR/LF stripped) are used as the value. This is the same convention used by the official Postgres, MariaDB, and Redis container images, so credentials no longer have to appear in plain `environment:` blocks where `docker inspect`, container logs, and `/proc/<pid>/environ` would expose them.
--- a/docs/source/usage.md
+++ b/docs/source/usage.md
@@ -367,6 +367,52 @@ The full set of configuration options are:
     `%` characters must be escaped with another `%` character,
     so use `%%` wherever a `%` character is used.
     :::
+- `postgresql`
+  - `host` - str: The PostgreSQL server hostname or IP address.
+    Required unless `connection_string` is provided.
+  - `port` - int: The PostgreSQL server port (Default: `5432`)
+  - `user` - str: The database user name (Optional)
+  - `password` - str: The database user password (Optional)
+  - `database` - str: The database name (Optional)
+  - `connection_string` - str: A full libpq connection string or URI
+    (e.g. `postgresql://user:pass@host/dbname`). When provided,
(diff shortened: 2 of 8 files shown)


seanthegeek (dismissed):

seanthegeek:
Pushed a rework to the branch (maintainer edit) to get this merge-ready for 10.1.0. Rebuilt cleanly on current `master` rather than replaying the branch, since it conflicted from the first commit and most touched lines were being rewritten anyway. Author attribution preserved (@fsvm88).

**Blockers addressed**
- **psycopg is now optional.** Moved to a `[postgresql]` extra (`pip install parsedmarc[postgresql]`); the import in `postgres.py` is guarded so `import parsedmarc` and `parsedmarc --help` still work without it, and `PostgreSQLClient` raises a clear install hint if constructed without the driver. Binary wheels aren't published for every platform, so it can't be a mandatory dependency.
- **forensic → failure rename** to match the project-wide rename (#659): table `dmarc_failure_report` / `dmarc_failure_sample_address`, `save_failure_report_to_postgresql`, the Grafana dashboard, and the docs. Config section stays `[postgresql]`.

**Also fixed / added**
- Failure-report de-duplication via a query-then-insert `IS NOT DISTINCT FROM` check on `(arrival_date_utc, reported_domain, source_ip_address, sample_subject)`, mirroring the Elasticsearch backend (aggregate and SMTP-TLS already use `ON CONFLICT`). Open to a hard `UNIQUE` constraint or different key columns if preferred — I avoided one only to dodge a PG15-only `NULLS NOT DISTINCT` dependency.
- Config parse raises `ConfigurationError` (not `exit(-1)`) on a missing `host`/`connection_string`.
- `PostgreSQLClient.close()` for clean CLI shutdown, plus a comment explaining why the two timestamp helpers must stay distinct (report begin/end are local time; record and SMTP-TLS dates are UTC).
- Tests moved to `tests/test_postgres.py` (helpers, mocked-DB save assertions on the SQL and bound parameters, a real-sample round trip, the install-hint guard, failure dedup, and close/reconnect) plus `TestParseConfigPostgreSQL` and `TestPostgreSQLCliWiring` in `tests/test_cli.py`. `tests.py` is no longer used (it was split in #774).
- CHANGELOG entry under a new `## 10.1.0` section. I left `constants.py` at `10.0.0` — the version bump is yours to make at release time.

Full suite: 607 passed; `ruff check`/`ruff format --check` clean; docs build succeeds.

_— written by Claude on Sean's behalf._

seanthegeek:
Follow-up push addressing the review points:

- **Failure dedup now mirrors Elasticsearch exactly** — keyed on arrival date + From + To + Subject (dropped the source-IP/domain key). NULL-safe via `IS NOT DISTINCT FROM`, and JSONB equality is semantic so From/To object key order doesn't matter. Verified against a real PostgreSQL 16: the two LinkedIn samples (`.eml` + `.crlf.eml`, same logical report) dedupe correctly, and a reprocess pass dedupes all four.
- **RFC 9990 / DMARCbis aggregate fields added to the schema** — `np`, `testing`, `discovery_method`, `generator`, `xml_namespace` on `dmarc_aggregate_report`, and `human_result` on the DKIM/SPF auth-result tables. Confirmed they populate from the RFC 9990 sample reports.
- **Coverage** — `codecov/patch` was failing (80.7%). Added tests for `create_tables`, connection-string/connect-failure handling, save-method error wrapping, the single-dict recipient path, the DMARCbis field mapping, and the CLI save/error-handling branches for all three report types. `parsedmarc/postgres.py` is now at 99% line coverage; the only gap is `_main`'s output-client-init retry path (the orchestration path the project already documents as out of scope).
- **CHANGELOG** — moved the PostgreSQL entry into the 10.0.0 section.

Full suite green (618 tests); `ruff check` / `ruff format --check` clean. The new tests run with `psycopg` absent (mocked at the SDK boundary, matching CI), and I separately validated the whole backend — schema, all saves, both dedup mechanisms, timezone handling, arrays/JSONB/INET — against a live PostgreSQL 16.

_— written by Claude on Sean's behalf._

Issue #581, opened 2024-12-01
Parameter since is not working correctly


dwydler:
Hello Sean,
I updated parsedmarc to version 8.16.0. Then added the parameter `since` to the parsedmarc.ini configuration file. The DMARC reports are retrieved from an email mailbox via IMAPS.

```
[general]
save_aggregate = True
save_forensic = True
save_smtp_tls = True
log_file = /var/log/parsedmarc/parsedmarc.log
silent = False
debug = True
delete = False

[mailbox]
reports_folder = INBOX
watch = True
test = True
check_timeout = 3600
#batch_size = 0
since = 1d
```

The email inbox contains emails/reports that are older than one day.
![image](https://github.com/user-attachments/assets/81aeeb2c-be3c-46d5-866a-516c323a408a)

After starting parsedmarc, no emails are read at first. In the same second, 8 emails are recognized. This corresponds to the number of emails in the screenshot.
```
2024-12-01 10:21:27,247 - INFO - [cli.py:1205] - Starting parsedmarc
2024-12-01 10:21:27,642 - DEBUG - [__init__.py:1593] - Only days and weeks values in 'since' option are                          considered for IMAP conections. Examples: 2d or 1w
2024-12-01 10:21:27,657 - DEBUG - [__init__.py:1612] - Found 0 messages in INBOX
2024-12-01 10:21:27,657 - DEBUG - [__init__.py:1619] - Processing 0 messages
2024-12-01 10:21:27,670 - INFO - [cli.py:1588] - Watching for email - Quit with ctrl-c
2024-12-01 10:21:27,909 - DEBUG - [__init__.py:1612] - Found 8 messages in INBOX
2024-12-01 10:21:27,909 - DEBUG - [__init__.py:1619] - Processing 8 messages
2024-12-01 10:21:27,909 - DEBUG - [__init__.py:1623] - Processing message 1 of 8: UID 364
2024-12-01 10:21:27,925 - INFO - [__init__.py:1199] - Parsing mail from noreply-dmarc-support@google.com on 2024-11-27 15:59:59-08:00
2024-12-01 10:21:28,081 - DEBUG - [utils.py:347] - Trying to fetch reverse DNS map from https://raw.githubusercontent.com/domainaware/parsedmarc/master/parsedmarc/resources/maps/base_reverse_dns_map.csv...
2024-12-01 10:21:28,122 - DEBUG - [utils.py:439] - IP address 2a01:111:f403:c20c::4 added to cache
[...]

PR #670, opened 2026-03-09
Pass mailbox since filter through watch_inbox callback

mkilijanek:
## Summary
- pass `since` through `watch_inbox()` into `get_dmarc_reports_from_mailbox()`
- wire CLI watch mode to pass `opts.mailbox_since` into `watch_inbox(...)`
- add regression test ensuring `since` is forwarded

## Testing
- `pytest tests.py -q`


seanthegeek:
Looks like this finally fixes https://github.com/domainaware/parsedmarc/issues/581

Thanks!

PR #727, opened 2026-04-25
Precompute SMTP TLS session-count totals on the report doc

seanthegeek:
## Why

The OpenSearch / Elasticsearch SMTP TLS report document maps `policies` as a `Nested` field, which is the right shape — each TLS-RPT policy is a self-contained unit. But under nested mapping, the per-policy `successful_session_count` and `failed_session_count` values can only be aggregated correctly via a Nested bucket aggregation, which the classic Visualize editor in OpenSearch Dashboards doesn't support cleanly.

The bundled dashboards had been working around this in ways that produced wrong numbers:

- **"SMTP TLS sessions"** (pie) used a `filters` bucket with KQL filters `policies.successful_session_count > 0` / `policies.failed_session_count > 0` and a `count` metric. Each slice was the number of *reports* in which at least one policy met the condition — not sessions.
- **"TLSRPT policies"** (pie) used a terms agg on `policies.policy_type.keyword` without a nested wrapper and a `count` metric. Each slice was approximately the number of *reports* containing at least one policy of that type — not policies.
- **"TLSRPT policies by domain"** (data table) used terms aggs on `policies.policy_domain.keyword` + `policies.policy_type.keyword` with a `count` metric. The value column was *reports* per (domain, type) — not sessions.

## What

Add six precomputed top-level fields to `_SMTPTLSReportDoc` (mirrored in both `parsedmarc/opensearch.py` and `parsedmarc/elastic.py`):

- `total_successful_session_count` — sum of per-policy `successful_session_count`
- `total_failed_session_count` — sum of per-policy `failed_session_count`
- `policy_count` — `len(policies)`
- `sts_policy_count`, `tlsa_policy_count`, `no_policy_found_policy_count` — per-RFC-8460-type counts

A shared `_smtp_tls_policy_totals()` helper computes from the input report's `policies[]` (source-of-truth dict, not the doc-class state). Missing / `None` count values are treated as 0; an empty `policies[]` produces all zeros; unknown `policy_type` values still bump `policy_count` but no per-type counter, and emit a debug log line.

Update the bundled OpenSearch dashboards NDJSON (`opensearch/opensearch_dashboards.ndjson`) to use the new fields:

- "SMTP TLS sessions" (pie → `metric`): two `Sum` aggregations with custom labels "Successful sessions" and "Failed sessions"
- "TLSRPT policies" (pie → `metric`): three `Sum` aggregations with custom labels "STS", "TLSA", "no-policy-found"
- "TLSRPT policies by domain" (data table): bucketing unchanged; metric swapped from `count` to `Sum(total_successful_session_count)` + `Sum(total_failed_session_count)`

I picked the single-vis multi-metric `metric` shape over the alternative (horizontal bar with a `filters` bucket + multiple `Sum` metrics, or two side-by-side `metric` panels) because it keeps every dashboard reference and panel layout unchanged — the visState swap is the entire diff for those vizes.

"SMTP TLS failures" is unchanged. Its bucketing is fundamentally per-report-per-failure-combination and the new totals don't help; the prompt explicitly scoped it out.

## OSD compatibility

The bundled NDJSON exports were already produced from OSD 3.5 (panel `version: "3.5.0"`) and use the legacy Kibana `kibanaSavedObjectMeta` saved-object schema that OSD 3.5 / 3.6 still accept. This change preserves those names and `migrationVersion` values verbatim — only `visState` payloads (and the index-pattern `fields` array) are modified, both of which remain stringified JSON in the same shape OSD already wrote. No new schema elements are introduced.

## Migration note

Existing TLS-RPT indices won't have the new fields populated. The updated vizes will show zeros for historical data until either reindexed or written-fresh. No automated backfill migration is included; that's a separate concern if anyone needs it.

## Tests

Seven new unit tests in `tests.py` cover the totals helper:

- a normal multi-policy report
- an empty `policies[]`
- a report with policies missing a count field
- all-zero counts
- all three known policy types in one report
- an unknown policy type (asserts the debug log line)
- the Elasticsearch helper matches the OpenSearch one for a representative input

Full test suite passes (168/168). `ruff check` and `ruff format --check` clean.

## Follow-up

The `dmarc-msp` repo carries an identical copy of `opensearch_dashboards.ndjson` and will need the same change applied as a separate PR.

## Test plan

- [x] `pytest tests.py` (168 passed)
- [x] `ruff check .` clean
- [x] `ruff format --check .` clean
- [ ] Import the modified NDJSON into a clean OSD 3.5 instance, open the SMTP TLS reporting dashboard, confirm the affected vizes render and show plausible numbers
- [ ] Same import test against OSD 3.6


seanthegeek:
I found a much better way

PR #729, opened 2026-04-26
Bulk reverse-DNS map coverage: top-500 ASN audit + KU re-research + curl fallback

seanthegeek:
## Summary

Three coordinated changes that lift the reverse-DNS map's IPv4 coverage from **85.10%** to **88.89%** of the bundled IPinfo Lite MMDB (+3.79 percentage points, ~118.2M additional IPv4 addresses now classified):

- **`collect_domain_info.py` — curl fallback for blocked/broken fetches.** Triggers on requests-side errors or non-2xx responses, runs `curl -kL --max-time N -A <browser-UA>` against the same scheme, parses `<head>` the same way as the primary path. Recovers six of eight cert-error domains in smoke testing. Trigger conditions and error reporting are documented inline.

- **Top-500 unmapped ASN-domain audit.** Walked every record in `ipinfo_lite.mmdb` to find `as_domain` values not yet in the map, took the top 500 by routed IPv4 count (each ≥ ~/15), enriched with `collect_domain_info.py`, and classified under the two-corroborating-sources rule. Yield: **435 newly-classified map entries** + 65 known-unknown additions.

- **Full re-research of the existing 3,606-entry KU file using the new curl fallback.** Recovered homepage content for 1,686 of 3,670 (45.9%) previously dark domains. Of those, 770 had a corroborating WHOIS or `as_name` alongside, and **508 cleared the strict service-category test** and moved out of known-unknown into the map. The remaining 262 recovered titles were brand-only / login-portal / under-construction pages where service category could not be assigned with confidence. Also removed a pre-existing `#name?` Excel auto-correction artifact.

## Coverage delta

| Metric | Before | After | Delta |
|---|---|---|---|
| `base_reverse_dns_map.csv` rows | 3,946 | **4,889** | +943 (+23.9%) |
| `known_unknown_base_reverse_dns.txt` lines | 3,606 | **3,162** | −444 net |
| MMDB IPv4 covered by map | 2,653,556,992 (85.10%) | **2,771,754,240 (88.89%)** | +118.2M IPv4 (+3.79 pp) |
| Distinct `as_domain` values still unmapped | 62,578 | 61,784 | −794 |

## Classification rules applied

[...]

PR #730, opened 2026-04-26
Map aliases for redirect targets + CC BY-SA 4.0 attribution

seanthegeek:
Two follow-ups to PR #729:

## 1. License attribution (`README.md`)

`base_reverse_dns_map.csv` is now a curated derivative of the bundled IPinfo Lite MMDB — many entries were generated by walking the database's `as_domain` / `as_name` fields for unmapped operators. IPinfo Lite ships under [Creative Commons Attribution-ShareAlike 4.0](https://creativecommons.org/licenses/by-sa/4.0/), which propagates to derivative works. Adds a `### License` subsection under the `## base_reverse_dns_map.csv` heading declaring the CSV under **CC BY-SA 4.0** with attribution to IPinfo for the underlying network identification data. Also updates the file-size estimate from "over 1,400" to "over 5,000" to reflect the current state.

## 2. Redirect-target aliases (`base_reverse_dns_map.csv` + `AGENTS.md`)

When a mapped domain's homepage redirects to a different host *for the same operator* (acquisition target's site, or a TLD/subdomain variant), PTR reverse-DNS reports observed in the wild may reference either domain. Mapping only the original loses attribution for the redirect target.

**Adds 91 aliases** discovered by sweeping the prior PR's classification artifacts. Each is a redirect target where:
- the original domain was newly mapped in PR #729,
- the redirect target's base wasn't already in the map,
- the target represents the same operator (acquisition or TLD/subdomain variant per AGENTS.md step 6),
- the target wasn't a placeholder / bot redirect / TLD-eTLD over-reduction.

Examples: `apogee.us` + `boldyn.com` both → Boldyn ISP; `sungardas.com` + `1111systems.com` both → 11:11 Systems MSP; `vodafone.is` + `syn.is` both → Sýn ISP; `sendinblue.com` + `brevo.com` both → Brevo (Sendinblue) Marketing; `tigo.com` + `millicom.com` both → Tigo ISP; `rockwellcollins.com` + `collinsaerospace.com` both → Collins Aerospace Defense; `zoom.us` + `zoom.com` both → Zoom SaaS.

**Codifies the practice in AGENTS.md** as a new paragraph under step 6 (the homepage-redirect disambiguation rule). The guardrail makes the rule symmetric with what's already there:
- Alias *only* for case 1 (acquisition) and case 3 (TLD/subdomain variant) where the redirect target is the same operator.
- *Do not* alias for case 2 (sister brand / shared infra) where the redirect target is a genuinely different operator. Cited example: do not alias `ziggo.nl` to UPC after the `chello.sk` fix.
- Skip generic-placeholder, bot-management, and TLD/eTLD redirect targets.
- When in doubt, drop the alias. A missing alias is recoverable; a wrong one mis-attributes mail.

**Also fixes four canonical-naming inconsistencies** the alias sweep surfaced — recent additions that didn't match pre-existing canonical brand names:
- `ga.gov`: "Georgia Government" → "State of Georgia" (matches `georgia.gov`)
- `goco.ca`, `radiant.net`: "Telus" → "TELUS" (matches `telus.com`)
- `vee.com.tw`: "VeeTime" → "VeeTIME" (matches `veetime.com`)

## Coverage

Aliases don't change MMDB IPv4 coverage (the targets weren't `as_domain`s). The win is PTR-side attribution coverage when DMARC reports reference the new operator's domain.

## Test plan

- [x] All 91 alias rows have valid `type` values from `base_reverse_dns_types.txt`
- [x] No collisions with existing map keys
- [x] `sortlists.py` ran cleanly (sort + dedupe + type-validation)
- [x] CRLF preserved
- [x] Four canonical-naming fixes validated against the pre-existing entries they're aligning to
- [ ] Reviewer: confirm the alias decisions, particularly the acquisition-driven ones (`apogee.us → Boldyn`, `sungardas.com → 11:11 Systems`, `eatel.com → REV`, `belairinternet.com → Zentro Internet`, `performive.com → CloudFirst`, `attachmate.com → Rocket Software`, `webpass.net → Google Fiber`, `welcomeitalia.it → Vianova`)
- [ ] Reviewer: confirm the AGENTS.md rule's caveat list is complete (sister-brand exclusion, placeholder/bot list, TLD over-reductions)


seanthegeek:
## Follow-up — sweep the rest of the map for missing redirect-target aliases

The 91 aliases in this PR cover only the domains added in PR #729 (the previous bulk batch). The rest of `base_reverse_dns_map.csv` — roughly 4,000 rows that predate this session, plus the ~500 added directly in PR #729 — was never run through a redirect-mismatch sweep, so missing aliases of the same shape almost certainly exist there too.

**Likely category sizes (rough, untested estimates):**
- Acquisitions/rebrands where the legacy domain stayed in the map but the acquirer's site is missing as an alias
- TLD/subdomain variants where the entry uses one form (`x.net`) and the live homepage redirects to another (`x.com`)
- Old corporate-rename cases (e.g. parent rebrands)

**To run a full sweep:**
1. Build a domain-only input from the entire current map (~5,489 rows)
2. Run `collect_domain_info.py` against it (curl fallback enabled). Wall-clock ~25–30 min at 16 workers.
3. Filter rows where `final_url`'s base host differs from the domain being classified, applying the same skip-list from the AGENTS.md step 6 codification (TLDs, placeholders, bot redirects, sister brands).
4. For each surviving candidate, propose the alias (matching the existing entry's `name`/`type`).
5. Review and apply.

This is bounded work — the same pattern as the alias sweep in this PR, just at 7× the input size. Estimated yield is probably 100–300 additional aliases (the existing map was hand-curated over years and presumably already has many of the obvious aliases, but acquisitions in the last 1–2 years would be the highest-yield gap).

Filing this as a comment rather than a separate issue so the context lives with the PR that introduced the rule. Happy to run it as a follow-up branch when there's bandwidth.

PR #770, opened 2026-05-15
Add support for Elastic Cloud Serverless projects

dvbnl:
Add support for Elastic Cloud Serverless projects. Setting serverless = True in the [elasticsearch] config section makes parsedmarc compatible with Elastic Serverless deployments, which previously rejected every write with an HTTP 400 error. Existing non-Serverless setups are unaffected.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,10 @@
 
 ### Enhancements
 
+#### Elastic Cloud Serverless compatibility
+
+New `[elasticsearch] serverless` config flag (env var `PARSEDMARC_ELASTICSEARCH_SERVERLESS`). Elastic Cloud Serverless manages sharding and replication itself and rejects the `number_of_shards` / `number_of_replicas` index settings with HTTP 400 — previously every write into a Serverless project failed at index-creation time. With the flag set, `create_indexes` strips those two keys from the settings sent to Elasticsearch and passes any other settings (e.g. `refresh_interval`) through unchanged. Non-Serverless deployments are unaffected.
+
 #### Docker-secret support via `_FILE` env vars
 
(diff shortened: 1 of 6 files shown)


seanthegeek:
Pushed some changes to the branch (maintainer edit) to address the review and get this merge-ready. Summary:

**Correctness fix (Copilot's first comment was right):** `create_indexes` previously skipped the entire `index.settings()` call in Serverless mode, which would silently drop any non-shard setting too. It now *filters out* just the Serverless-rejected keys (`number_of_shards` / `number_of_replicas`, held in a `_SERVERLESS_REJECTED_SETTINGS` frozenset) and passes everything else (e.g. `refresh_interval`) through unchanged. The existing `test_creates_with_custom_settings` already demonstrated non-shard settings flow through this path, so this keeps that contract intact under Serverless.

**Tests added (Copilot's third comment):**
- `tests/test_elastic.py::TestCreateIndexesServerless` — asserts shard/replica keys are stripped while other keys pass through, and that `index.settings()` is skipped when only rejected keys remain. `_SERVERLESS` is saved/restored in setUp/tearDown.
- `tests/test_cli.py` — a parse test for the `serverless` flag plus an end-to-end guard asserting `elastic.set_hosts(..., serverless=True)` is actually reached, so the config-parse and the plumbing can't drift apart.

**Smaller items:** dropped a redundant `bool()` wrap around `getboolean` in `cli.py`, added a CHANGELOG entry, and reworded the docs to describe the filter-not-skip behavior.

**On Copilot's second comment (module-level `_SERVERLESS` global):** I kept the module-global rather than threading `serverless` through every `save_*_to_elasticsearch` call. The rationale: this module already relies on module-global connection state via `connections.create_connection()` in `set_hosts()` — every save path reads back that single default connection. parsedmarc runs as a single-process CLI/watcher with one ES connection per process, so `_SERVERLESS` mirroring that existing pattern is internally consistent, and refactoring it out would be churn well beyond this PR's scope. I documented the global (and why it's safe here) with a comment at the top of the module. If we ever want to support multiple concurrent ES connections in one process, that's a separate refactor that should move *both* the connection and this flag off module state together.

All 566 tests pass; `ruff check` and `ruff format --check` are clean.

_— written by Claude on Sean's behalf._

Issue #811, opened 2026-07-08
Timestamp handling is timezone-dependent: failure-report timestamps skewed by host UTC offset in ES/OpenSearch/Splunk; JSON/CSV dates are ambiguous local time


mkilijanek:
### Bug 1 (concrete, verified): failure-report timestamps are off by the host's UTC offset in three output sinks

`parse_failure_report()` generates `arrival_date_utc` as a **UTC** wall-clock string (`parsedmarc/__init__.py:1493-1497` — aware-UTC datetime → `strftime("%Y-%m-%d %H:%M:%S")`, offset discarded).

The consumers then parse it back **naive** and convert with `.timestamp()`:

- `parsedmarc/elastic.py:663-664`
  ```python
  arrival_date = human_timestamp_to_datetime(failure_report["arrival_date_utc"])  # naive
  arrival_date_epoch_milliseconds = int(arrival_date.timestamp() * 1000)          # naive -> assumed LOCAL
  ```
- `parsedmarc/opensearch.py:663-664` — identical code
- `parsedmarc/splunk.py:166` — `human_timestamp_to_unix_timestamp(report["arrival_date_utc"])`, which also ends in a naive `.timestamp()`

Per the Python docs, a naive `datetime.timestamp()` "assume[s] that it represents local time" (<https://docs.python.org/3/library/datetime.html#datetime.datetime.timestamp>). So a UTC string is re-interpreted as local wall-clock, and the resulting epoch is wrong by the host's UTC offset on every non-UTC machine.

Verified live on a `TZ=Europe/Warsaw` host:

```python
>>> naive = human_timestamp_to_datetime("2024-01-15 12:00:00")  # a UTC string, parsed naive
>>> int(naive.timestamp()) - 1705320000   # true epoch for 12:00 UTC
-3600                                      # 1 h skew (2 h in summer, DST)
```

Impact: the skewed epoch is **stored** as the `arrival_date` field of every failure-report document in Elasticsearch/OpenSearch (`elastic.py:765`, `opensearch.py:763`), used in the duplicate-detection `match` query (`elastic.py:675`), and used as the HEC event `time` for every failure report sent to Splunk. Dashboards on non-UTC collectors show failure events 1–2 h off; dedup can miss across hosts in different timezones.

The fix is small: these three call sites know the string is UTC, so attach `timezone.utc` when parsing (or keep the aware datetime from the parser instead of round-tripping through a string).

[...]

PR #812, opened 2026-07-08
Fix failure-report timestamp skew on non-UTC hosts in ES/OpenSearch/Splunk outputs

mkilijanek:
Fixes bug 1 of #811 (the concrete `arrival_date_utc` skew). Bug 2 of that issue — the local-time strings in JSON/CSV output — is intentionally **not** touched here, since it changes output values and needs a maintainer decision first.

## The bug

`parse_failure_report()` generates `arrival_date_utc` as a **UTC** wall-clock string (aware-UTC datetime → `strftime`, offset discarded). Three output sinks then parsed it back into a *naive* `datetime` and called `.timestamp()`:

- `parsedmarc/elastic.py` — epoch-ms stored as the failure document's `arrival_date` field and used in the dedup `match` query
- `parsedmarc/opensearch.py` — identical code
- `parsedmarc/splunk.py` — the HEC event `time` via `human_timestamp_to_unix_timestamp()`

Per the Python docs, a naive `datetime.timestamp()` "assume[s] that it represents local time" (<https://docs.python.org/3/library/datetime.html#datetime.datetime.timestamp>), so the UTC string was re-interpreted as local wall-clock and the resulting epoch came out wrong by the host's UTC offset.

## Who is affected

Any deployment whose host timezone is not UTC. The skew equals the negated host UTC offset at the arrival date, so both magnitude and direction vary by region:

| Host timezone | UTC offset | Stored `arrival_date` / HEC `time` skew |
|---|---|---|
| UTC (typical servers, Docker containers without `TZ`) | 0 | none — bug invisible |
| Central Europe (Berlin, Warsaw, Paris) | +1 / +2 (DST) | −1 h winter / −2 h summer |
| US Eastern | −5 / −4 (DST) | **+5 h / +4 h** |
| India | +5:30 | −5 h 30 min |
| Eastern Australia (Sydney) | +10 / +11 (DST) | −10 h / −11 h — enough to land events on the wrong day in daily dashboards |

Three factors kept this hidden: CI and most server/container deployments run UTC (no skew there); it only affects failure/RUF reports, which are rare compared to aggregates; and a 1–2 h dashboard offset is easy to misattribute to a Kibana/Grafana display timezone. Additional wrinkles: DST changes the skew twice a year, making stored history internally inconsistent within a single installation, and two collectors in different timezones compute different epochs for the same report, so the dedup query fails to match and the report is double-indexed.

## The fix

`human_timestamp_to_datetime()` / `human_timestamp_to_unix_timestamp()` gain an `assume_utc` keyword that attaches `timezone.utc` to naive parses (timestamps carrying an explicit offset are unaffected). The three `arrival_date_utc` call sites pass `assume_utc=True`; every other caller keeps the existing local-time semantics, whose round-trip with `timestamp_to_human` stays self-consistent on a single host.

No index-naming change: the failure index date partition was already derived from the UTC wall-clock fields of the parsed datetime, which are identical before and after this fix.

## Test plan

- [x] 7 new regression tests: 4 in `tests/test_utils.py` for the helper semantics (aware-UTC result, documented local-time default, unix-timestamp variant, explicit offsets not overridden), and 1 per sink asserting the *actual value sent* — the epoch-ms passed into `_FailureReportDoc` and the HEC event `time` — equals the true UTC epoch.
- [x] The sink tests force `TZ=Europe/Warsaw` via `time.tzset()` (with cleanup and a POSIX skip guard), so they catch the skew even on UTC CI runners; verified all 3 **fail** against the unfixed code and pass with the fix.
- [x] Full suite passes on a UTC-equivalent environment; on a non-UTC dev machine the only failures are the 4 pre-existing aggregate index-name tests tracked as bug 2 of #811 (unchanged by this PR).
- [x] `ruff check` / `ruff format --check` / `pyright` clean.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,11 @@
 # Changelog
 
+## Unreleased
+
+### Bug fixes
+
+- **Failure-report timestamps are no longer skewed by the host's UTC offset in the Elasticsearch, OpenSearch, and Splunk HEC outputs** ([#811](https://github.com/domainaware/parsedmarc/issues/811), bug 1). `arrival_date_utc` is a UTC wall-clock string, but the three sinks parsed it into a naive `datetime` and called `.timestamp()`, which per the Python docs interprets naive values as *local* time — so on any non-UTC host, the epoch stored as the ES/OpenSearch `arrival_date` field, used in the failure-report dedup query, and sent as the Splunk HEC event `time` was off by the host's UTC offset (1–2 h for most of Europe). `human_timestamp_to_datetime()` / `human_timestamp_to_unix_timestamp()` gained an `assume_utc` keyword that attaches `timezone.utc` to naive parses, and the `arrival_date_utc` consumers now use it.
+
 ## 10.2.0
 
 ### Changes
--- a/parsedmarc/elastic.py
+++ b/parsedmarc/elastic.py
@@ -660,7 +660,12 @@ def save_failure_report_to_elasticsearch(
     for original_header in original_headers:
         headers[original_header.lower()] = original_headers[original_header]
 
-    arrival_date = human_timestamp_to_datetime(failure_report["arrival_date_utc"])
+    # arrival_date_utc is a UTC wall-clock string; without assume_utc the
+    # naive .timestamp() below would interpret it as local time and skew
+    # the epoch by the host's UTC offset.
+    arrival_date = human_timestamp_to_datetime(
+        failure_report["arrival_date_utc"], assume_utc=True
+    )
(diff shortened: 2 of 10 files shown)


seanthegeek:
Pushed a small follow-up commit (3a2d0c3) directly to the branch: extracted the repeated TZ-forcing setup/cleanup from the four regression tests into a shared `tests/tzutil.py` `force_tz()` helper, and fixed the `human_timestamp_to_unix_timestamp()` docstring (`float` → `int`). Verified the refactored tests still fail against the unfixed code and the full suite + ruff + pyright stay clean.

Issue #137, opened 2020-01-08
MaxMind registration and licensing changes


JamesWhyley:
Hello,

Firstly, I very much enjoy ParseDMARC, it's an excellent little tool. 

Secondly, as I'm sure people may have noticed, MaxMind changed their licensing arrangements on December 30th to require registration as per https://blog.maxmind.com/2019/12/18/significant-changes-to-accessing-and-using-geolite2-databases/ and this now causes ParseDMARC to error if you're not running it offline. 

Observed behaviour, using the free DB offering, is that it does the usual call to the old GeoLite2 download location, finds nothing, retries, throws an error, then moves on to the next message. It would appear that nothing subsequently appears in elasticsearch either.

Log looks llike this `   DEBUG:__init__.py:1074:Processing message 1 of 314: UID 74051
   ERROR:utils.py:331:Error downloading https://geolite.maxmind.com/download/geoip/database/GeoLite2-Country.tar.gz: HTTPSConnectionPool(host='geolite.maxmind.com', port=443): Max retries exceeded with url: /download/geoip/database/GeoLite2-Country.tar.gz (Caused by NewConnectionError('<urllib3.connection.VerifiedHTTPSConnection object at 0x7fd10518df10>: Failed to establish a new connection: [Errno -2] Name or service not known'))`

Processing offline works as usual.

Edit: Sorry, I should have said it works perfectly fine if you have a pre-existing GeoLite2 db as well.

robertitox:
Dear, I have a Debian 10 box with parsedmarc and I know yoy said. I've downloaded the GeoLite2-Country.mmdb and put it in /usr/share/GeoIP. After that I ran "parsedmarc -c /etc/parsedmarc.ini" and the error you mentioned has disappeared.

But my problem is that I can see the Map of Message source countries at all on Kibana, please can anybody tell me how to see the map locations ?

I've just setup "chown -R root.parsedmarc /usr/share/GeoIP" and "chmod -R 664 /usr/share/GeoIP".

Nw inside /usr/share/GeoIP I have these files:

GeoIPASNum.dat  
GeoIPCity.dat  
GeoIP.dat  
GeoIPv6.dat  
GeoLite2-Country.mmdb

Thanks a lot !

dragoangel:
Need ad some API keys to config and use new links to download GeoIP

seanthegeek:
@robertitox What do you mean? The map and source data should always be included in the dashboard.

seanthegeek:
[...]

PR #813, opened 2026-07-08
Stop system GeoIP files from shadowing the bundled IPinfo database

### LINKED ISSUE #810: System GeoLite2-Country.mmdb silently shadows the bundled IPinfo database, disabling ASN enrichment
### Summary

`_get_ip_database_path()` in `parsedmarc/utils.py` searches a list of well-known system database paths **before** falling back to `_IP_DB_PATH` (the database selected by `load_ip_db()` — downloaded, cached, or bundled IPinfo Lite):

```python
# parsedmarc/utils.py (master), _get_ip_database_path()
db_paths = [
    "ipinfo_lite.mmdb",
    "GeoLite2-Country.mmdb",
    "/usr/local/share/GeoIP/GeoLite2-Country.mmdb",
    "/usr/share/GeoIP/GeoLite2-Country.mmdb",
    "/var/lib/GeoIP/GeoLite2-Country.mmdb",
    ...
]
if db_path is None:
    for system_path in db_paths:
        if os.path.exists(system_path):
            db_path = system_path
            break
if db_path is None:
    if _IP_DB_PATH is not None:
        db_path = _IP_DB_PATH        # <- never reached when a system file exists
```

On any host that has a distro GeoIP package installed (Fedora/Debian/Ubuntu commonly ship `/usr/share/GeoIP/GeoLite2-Country.mmdb`), every lookup silently uses the **country-only** system database instead of the IPinfo Lite database that `load_ip_db()` just downloaded/bundled. Two consequences:

1. **ASN enrichment is silently disabled.** GeoLite2-Country has no ASN data, so `asn`, `as_name`, and `as_domain` come back `None` for every IP. This also disables the ASN-fallback path into `base_reverse_dns_map.csv` (used when an IP has no PTR), so source attribution quality drops substantially — with no error, only the generic "IP database is more than a month old" warning if the system file is old.
2. **Country data comes from whatever vintage the system file is.** Verified on a Fedora workstation whose distro-installed GeoLite2-Country.mmdb is dated **December 2019** — those MaxMind-package files stopped updating for many users after the 2019 MaxMind licensing change (#137), so 5+ year old files are realistic in the wild.

### Reproduction (verified)

On a host with `/usr/share/GeoIP/GeoLite2-Country.mmdb` present:

```python
>>> from parsedmarc.utils import get_ip_address_info
>>> get_ip_address_info("8.8.8.8", offline=True)
# asn=None, as_name=None, as_domain=None, name=None
# expected (bundled ipinfo_lite.mmdb): asn=15169, as_name='Google LLC', as_domain='google.com'
```

The bundled `parsedmarc/resources/ipinfo/ipinfo_lite.mmdb` returns the full record for the same IP — it is simply never consulted because the system path matches first.

This is also why `tests/test_utils.py`'s MMDB tests (`testIpAddressInfoSurfacesASNFields`, `testIpAddressInfoFallsBackToASNMapEntryWhenNoPTR`) fail on such developer machines while CI stays green (GitHub runners have no GeoIP files).

### Suggested fix

Invert the precedence so the database parsedmarc manages wins over incidental system files:

1. explicit `db_path` argument (already first — unchanged)
2. `_IP_DB_PATH` set by `load_ip_db()` (downloaded → cached → bundled IPinfo Lite)
3. system GeoLite2/DBIP paths as a last resort only

Two smaller related nits in the same list: the leading relative entries (`"ipinfo_lite.mmdb"`, `"GeoLite2-Country.mmdb"`, `"dbip-country*.mmdb"`) make lookups CWD-dependent, and logging which file was actually opened (debug level) would make this class of problem diagnosable from a `--debug` run.

Happy to send a PR if the reordering direction is acceptable.

mkilijanek:
Fixes #810.

## The bug

`_get_ip_database_path()` in `parsedmarc/utils.py` searched a list of well-known system database paths — including `/usr/share/GeoIP/GeoLite2-Country.mmdb` and CWD-relative names like `GeoLite2-Country.mmdb` — **before** `_IP_DB_PATH`, the database selected by `load_ip_db()` (downloaded → cached → bundled IPinfo Lite).

On any host with a distro GeoIP package installed (Fedora/Debian/Ubuntu commonly ship one, often as an incidental dependency of an unrelated package), every lookup silently used the country-only system database instead of the IPinfo Lite database parsedmarc just downloaded or bundled:

- **ASN enrichment was silently disabled.** GeoLite2-Country carries no ASN data, so `asn`, `as_name`, and `as_domain` came back `None` for every IP — which also disabled the ASN-fallback path into `base_reverse_dns_map.csv` used when an IP has no PTR.
- **Country data came from whatever vintage the system file was.** Verified live on a Fedora workstation whose distro-installed `GeoLite2-Country.mmdb` dated to **December 2019** — realistic in the wild, since many of those package-managed files stopped updating after the 2019 MaxMind licensing change (#137).
- The only symptom was the generic "IP database is more than a month old" warning.

## The fix

New precedence, so the database parsedmarc manages always wins over incidental system files:

1. explicit `ip_db_path` (unchanged)
2. `_IP_DB_PATH` selected by `load_ip_db()` (downloaded → cached → bundled)
3. the bundled copy directly — covers library callers that never call `load_ip_db()`
4. system GeoLite2/DBIP paths as a true last resort, only consulted when the bundled data file is missing

The selected database file is now logged at debug level (`Using IP database at <path>`), so a `--debug` run shows which file answered — this class of problem was previously undiagnosable from logs.

## Docs / behavior change

The automatic system-path pickup was **documented behavior** ("Using MaxMind GeoLite2" in `installation.md` said parsedmarc "will pick up" a GeoLite2 database from standard paths). That section now instructs MaxMind users to point `ip_db_path` at the file explicitly:

```ini
[general]
ip_db_path = /usr/share/GeoIP/GeoLite2-Country.mmdb
```

with a migration note explaining why the automatic pickup was removed. The `CHANGELOG.md` entry carries the same warning for users upgrading.

## Test plan

- [x] 2 new regression tests in `tests/test_utils.py` that reproduce the shadowing **portably** (CI runners have no `/usr/share/GeoIP`): the fallback list contains CWD-relative names, so a decoy `GeoLite2-Country.mmdb` in a temp working directory triggers the same code path. One asserts the bundled database wins (full ASN record for 8.8.8.8); the other asserts `_IP_DB_PATH` wins via the new selection debug log. Both **fail on the unfixed code** (verified by stashing the source change).
- [x] Live verification on an affected host: the 3 MMDB-dependent tests in `tests/test_utils.py` (`testIpAddressInfoSurfacesASNFields`, `testIpAddressInfoFallsBackToASNMapEntryWhenNoPTR`, `testIPinfoAPIPrimarySourceAndInvalidKeyIsFatal`) failed on a Fedora workstation with a Dec-2019 system GeoLite2 file and pass with this fix; `get_ip_address_info("8.8.8.8", offline=True)` now returns `asn=15169 / Google LLC / google.com` instead of `None`s.
- [x] Full suite: no new failures.
- [x] `ruff check` / `ruff format --check` / `pyright` clean.


seanthegeek:
Rebased onto master (resolved the `CHANGELOG.md` conflict with #812's Unreleased section) and pushed a follow-up commit (5830c3d) addressing review feedback:

- Corrected the nothing-found-anywhere comment — the `os.stat()` age check raises `FileNotFoundError` before `open_database()` would.
- The `Using IP database at ...` debug line now logs once per selected path (tracked via `_LAST_LOGGED_IP_DB_PATH`) instead of on every uncached lookup, so `--debug` runs over large batches aren't flooded.
- Added coverage for the previously untested fallback tiers: system-path fallback when the bundled DB is missing, `FileNotFoundError` when nothing exists anywhere, the stale-database warning, and the log-once-per-path behavior. `_get_ip_database_path` is now fully covered.

Verified: full suite passes (681 + 32 subtests), the two core regression tests still fail against the unfixed code, ruff/pyright clean.

PR #815, opened 2026-07-08
Make Microsoft Graph connection activity observable

### LINKED ISSUE #814: No visibility into Microsoft Graph connection activity — mailsuite/azure/msgraph logs are dropped even with --debug
### Summary

When the Microsoft Graph mailbox backend fails or misbehaves (e.g. Certificate auth against Exchange Online), `parsedmarc --debug` shows essentially nothing about the connection: no configuration summary, no token-acquisition activity, no HTTP requests. An operator cannot tell whether the problem is their parsedmarc config, a parsedmarc/mailsuite bug, or an Exchange Online / Entra ID configuration issue (missing API permission, application access policy, certificate mismatch…), because the evidence that would distinguish those — e.g. the `AADSTS` error codes returned by the token endpoint — is never surfaced.

### Root cause

parsedmarc's logging setup only attaches handlers and sets levels on its own logger (`parsedmarc.log`, see `cli.py` module-level handler + the `--verbose`/`--debug` level block in `_main`). Since the mailbox layer moved to mailsuite (#741), all Microsoft Graph activity logs under **different logger names** that parsedmarc never configures:

- `mailsuite.mailbox.graph` — connection/folder operations (`logger = logging.getLogger(__name__)` in mailsuite)
- `azure.identity` — credential and token acquisition, including `AADSTS` failures from `login.microsoftonline.com`
- `msgraph` / `httpx` — Graph SDK request/response activity

None of these have a handler, so everything below WARNING is silently dropped (WARNING+ only leaks through `logging.lastResort` unformatted). `--debug` has no effect on any of them.

Additionally, `_main()` calls `MSGraphConnection(...)` with no log line before or after, so a hang during credential construction or an interactive prompt leaves no trace at all.

### Proposed change (parsedmarc-side, no mailsuite changes required)

1. **Redacted connection summary at INFO before connecting** — auth method, tenant_id, client_id, mailbox, graph_url; DEBUG adds token_file, certificate_path, and set/unset flags for secrets. Secret *values* (`password`, `client_secret`, `certificate_password`) are never logged, and a regression test asserts they don't appear in captured output.
2. **Success log with elapsed time** after the connection object is initialized.
3. **Propagate parsedmarc's log level and handlers to the dependency loggers** (`mailsuite`, `azure`, `msgraph`, `httpx`, `httpcore`): follow the parsedmarc level when `--verbose`/`--debug` is set, and never quieter than WARNING at the default level (so today's stray warnings keep appearing, now formatted). This also benefits IMAP users, since `IMAPConnection` lives in mailsuite too.
4. Docs: a short troubleshooting note in the `msgraph` section of usage.md explaining what `--debug` now surfaces.

Deeper step-by-step auth logging inside the connection itself belongs in mailsuite and can be a follow-up there; the items above are all controllable from parsedmarc and are enough to distinguish config-side vs service-side failures (the azure.identity DEBUG output includes the token endpoint's error responses).

I have a PR ready for this.

### Environment

- parsedmarc 10.2.0, `[msgraph]` with `auth_method = Certificate` (client_id + tenant_id + certificate_path + certificate_password) against Exchange Online

mkilijanek:
Fixes #814.

## Problem

With the Microsoft Graph backend (e.g. `auth_method = Certificate` against Exchange Online), `parsedmarc --debug` shows essentially nothing about the connection. An operator troubleshooting a dead mailbox connection cannot tell whether the cause is their parsedmarc config, a parsedmarc/mailsuite bug, or an Exchange Online / Entra ID configuration issue, because:

- parsedmarc only configures its own logger (`parsedmarc.log`). Since the mailbox layer moved to mailsuite (#741), all Graph activity logs under logger names parsedmarc never touches — `mailsuite.mailbox.graph` (connection/folder operations), `azure.identity` (token acquisition, including the **`AADSTS` error codes** that pinpoint whether the tenant, app registration, certificate, or permissions are wrong), and `msgraph`/`httpx` (Graph SDK HTTP traffic). None of these have a handler, so everything below WARNING is silently dropped regardless of `--debug`.
- `_main()` calls `MSGraphConnection(...)` with no log line before or after, so a hang during credential construction leaves zero trace.

## Changes (all parsedmarc-side; no mailsuite changes needed)

1. **Redacted connection summary at INFO** before connecting:
   ```
   Connecting to Microsoft Graph (auth_method=Certificate, tenant_id=..., client_id=..., mailbox=shared@example.com, graph_url=https://graph.microsoft.com)
   ```
   plus a `--debug` detail line with `certificate_path`, `token_file`, and set/not-set flags for secrets. **Secret values (passwords, client secrets, certificate passwords) are never logged** — a regression test captures the full DEBUG output and asserts the configured secret string does not appear anywhere in it.
2. **Timing line** after the connection object is created (`Microsoft Graph connection initialized in 0.42 seconds`). A code comment notes that app-only methods acquire their first token lazily on the first mailbox call, so auth failures can still surface later.
3. **`_configure_dependency_logging()`**: parsedmarc's `--verbose`/`--debug` level now propagates to the `mailsuite`, `azure`, `msgraph`, `httpx`, and `httpcore` loggers, with handlers synced to exactly the parsedmarc logger's own handlers (console + optional log file) — so a SIGHUP reload that swaps the log file neither duplicates output nor keeps writing to a closed handler. At the default level, dependency loggers sit at WARNING: their warnings keep surfacing (now formatted, instead of leaking bare through `logging.lastResort`) without adding new noise. Called from `_main()` and from the SIGHUP reload path. This also benefits the IMAP backend, which lives in mailsuite too.

With this, `--debug` output for a misconfigured Certificate setup shows the azure-identity token request against `login.microsoftonline.com/<tenant>` and the exact `AADSTS` error (e.g. `AADSTS700027` certificate mismatch, `AADSTS7000215/500011` wrong client/tenant, missing application access policy errors from Graph itself) — enough to route the problem to the right team without adding print statements.

Docs: a Troubleshooting tip in the `msgraph` section of `usage.md` explains what `--verbose`/`--debug` now surface.

## Test plan

- [x] 4 new tests in `tests/test_cli.py` (`TestMSGraphConnectionLogging`): the INFO summary carries auth method/tenant/client/mailbox/URL and the timing line; **secret values never appear in DEBUG output**; `--debug` sets the five dependency loggers to DEBUG with parsedmarc's handlers attached; the default level leaves them at WARNING. Dependency-logger state is snapshotted and restored per-test so process-global loggers don't leak between tests.
- [x] All four fail on the unfixed code (verified by stashing the `cli.py` change).
- [x] Full suite: no new failures. `ruff check` / `ruff format --check` / `pyright` clean.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,13 @@
 
 ## Unreleased
 
+### Changes
+
+- **Microsoft Graph connections are now observable** ([#814](https://github.com/domainaware/parsedmarc/issues/814)). Previously `--debug` showed nothing about Graph connection activity: the mailbox layer logs under `mailsuite.mailbox.graph`, token acquisition under `azure.identity`, and HTTP traffic under `httpx`/`msgraph` — none of which parsedmarc configured, so everything below WARNING was silently dropped, and `_main()` logged nothing around the connection attempt itself. Now:
+  - A **redacted connection summary** is logged at INFO before connecting (auth method, tenant ID, client ID, mailbox, Graph URL), with a `--debug` detail line adding the certificate path, token-file path, and set/not-set flags for secrets. Secret values (passwords, client secrets, certificate passwords, client assertions) are never logged, and a regression test asserts they don't appear in captured output.
+  - A timing line (`Microsoft Graph connection initialized in N seconds`) is logged after the connection object is created.
+  - parsedmarc's `--verbose`/`--debug` level now **propagates to the dependency loggers** (`mailsuite`, `azure`, `msgraph`, `httpx`, `httpcore`), so `--debug` surfaces azure-identity's token-endpoint activity — including the `AADSTS` error codes that distinguish a local config problem from an Exchange Online / Entra ID one — and the Graph SDK's HTTP requests. At the default level, dependency loggers sit at WARNING so their warnings keep surfacing (now formatted) without new noise. This also benefits the IMAP backend, which lives in mailsuite too.
+
 ### Bug fixes
 
 - **Failure-report timestamps are no longer skewed by the host's UTC offset in the Elasticsearch, OpenSearch, and Splunk HEC outputs** ([#811](https://github.com/domainaware/parsedmarc/issues/811), bug 1). `arrival_date_utc` is a UTC wall-clock string, but the three sinks parsed it into a naive `datetime` and called `.timestamp()`, which per the Python docs interprets naive values as *local* time — so on any non-UTC host, the epoch stored as the ES/OpenSearch `arrival_date` field, used in the failure-report dedup query, and sent as the Splunk HEC event `time` was off by the host's UTC offset (1–2 h for most of Europe). `human_timestamp_to_datetime()` / `human_timestamp_to_unix_timestamp()` gained an `assume_utc` keyword that attaches `timezone.utc` to naive parses, and the `arrival_date_utc` consumers now use it.
--- a/docs/source/usage.md
+++ b/docs/source/usage.md
@@ -263,6 +263,18 @@ The full set of configuration options are:
     username, you must grant the app `Mail.ReadWrite.Shared`.
     :::
 
+    :::{tip}
+    **Troubleshooting connections.** Run with `--verbose` to log a
+    redacted connection summary (auth method, tenant, client ID,
+    mailbox, Graph URL) before the connection attempt, and with
+    `--debug` to additionally surface the underlying library activity —
+    `azure.identity` token acquisition (including `AADSTS` error codes
+    from Entra ID, which distinguish a local configuration problem from
+    an Exchange Online-side one), Microsoft Graph SDK requests, and
+    `httpx` HTTP request lines. Secret values (passwords, client
+    secrets, certificate passwords) are never written to logs.
+    :::
+
     :::{warning}
     If you are using the `ClientSecret` auth method, you need to
     grant the `Mail.ReadWrite` (application) permission to the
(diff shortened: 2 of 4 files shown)


seanthegeek:
Rebased onto master, resolving the three conflict points:

- `parsedmarc/cli.py`: dropped #809's simpler `Connecting to Microsoft Graph mailbox %s` INFO line in favor of this PR's redacted summary, and added a `client_assertion set/not set` flag to the `--debug` detail line since #809 introduced that secret after this branch forked.
- `tests/test_cli.py`: removed #809's `testCliLogsMsGraphConnectionAttempt` (strictly superseded by `testCliLogsMsGraphConnectionSummaryAndTiming`, which asserts the same INFO-before-connect behavior with richer field checks), and extended `testCliMsGraphLoggingNeverLogsSecretValues` to also run a `ClientAssertion` config and assert the assertion JWT value never appears in `--debug` output.
- `CHANGELOG.md`: merged this PR's `### Changes` entry into the existing Unreleased section ahead of the `### Bug fixes` entries from #812/#813, and added client assertions to the never-logged secrets list.

Verified after rebase: full suite 684 passed + 32 subtests, ruff and pyright clean.

seanthegeek:
Pushed 3130eb8 with the two review hardening items — one applied, one verified-and-declined:

- **`propagate = False`** is now set on the dependency loggers during handler sync, so a stray `logging.basicConfig()` anywhere in the process can't double-print every dependency record through the root logger. The test class snapshots/restores `propagate` alongside levels and handlers, and the enable test asserts it.
- **`kiota` was NOT added to `_DEPENDENCY_LOGGERS`**: checking the installed `kiota_http` / `kiota_authentication_azure` / `kiota_abstractions` packages showed they contain no Python `logging` usage at all — their observability is OpenTelemetry tracing — so there are no records to enable. A code comment now documents the deliberate absence so it doesn't get "fixed" later.

Full suite 684 passed + 32 subtests; ruff/pyright clean.

PR #823, opened 2026-07-14
Don't delete/archive mailbox messages until output sinks confirm the save

### LINKED ISSUE #242: Messages should only be moved after parsing and storing result is succesful
Currently messages are being moved away before the result is stored successfully, e.g. in Elasticsearch. This leads to report being marked as processed although there may be an error while storing the result.

mkilijanek:
## Summary

`get_dmarc_reports_from_mailbox()` archives or deletes every successfully-parsed message *before* the caller writes the parsed reports to any output sink (Elasticsearch/OpenSearch/Splunk/S3/Kafka/PostgreSQL/etc). If a sink fails, the report was never persisted anywhere and its source message was already gone from -- or moved out of -- the mailbox, with no way to recover it. A user reported exactly this after an Elasticsearch outage.

## Fix

`get_dmarc_reports_from_mailbox()` gains an optional keyword-only `save_callback` parameter, invoked once per fetched batch with that batch's newly parsed reports, after parsing but before the batch's messages are deleted/moved. The batch is only committed (dedup cache updated, messages archived/deleted) when the callback doesn't return `False` or raise; a failed batch is left in place for automatic retry on the next run or poll. `save_callback=None` (the default) preserves prior behavior for any batch that completes without an unhandled exception -- fully backward compatible for existing library callers.

`watch_inbox()`'s `callback` parameter is now passed through as `save_callback` instead of being invoked after archiving, so the same protection applies to continuous/`--watch` mode. The CLI wires this up via a `mailbox_save_callback` adapter built on `process_reports()`, which now returns its accumulated output errors instead of silently discarding them -- an output failure for a batch means that batch's messages stay in the mailbox, independent of `fail_on_output_error` (which only ever controlled the process exit code, not data safety).

This is deliberately an all-or-nothing rule per batch: archiving on partial sink success would still permanently lose data for whichever sink failed. Retrying can result in duplicate delivery to output destinations that don't deduplicate (Kafka, S3, Splunk HEC, Syslog, GELF, Webhook); Elasticsearch, OpenSearch, and PostgreSQL already dedupe via `AlreadySaved`.

Two related fixes found while implementing this:
- `save_output()` (the JSON/CSV/`.eml`-sample writer) was the only sink not wrapped in a try/except, so a disk-full or permission error there crashed the run uncaught instead of being recorded like every other sink's failures -- now caught and recorded consistently, and correctly blocks mailbox archiving like any other sink failure.
- `index_prefix_domain_map` filtering of SMTP TLS reports, previously applied as a side effect of a single `process_reports()` call covering everything, is now applied explicitly to the `parsing_results` dict fed to `email_results()` -- needed because mailbox-derived and file-derived reports are now saved via separate calls, so the filtering side effect no longer reaches the emailed summary automatically.

## Test plan

- [x] New tests in `tests/test_init.py` (`TestGetDmarcReportsFromMailboxMaildir`, real on-disk Maildir, no mocks) covering: failed-callback leaves messages in place, exception in callback leaves messages in place, successful callback still archives as before, delete-mode respects a failed callback, invalid-message handling is unaffected by the callback, and a failed-then-retried batch is not deduplicated against the poisoned dedup cache.
- [x] New/updated tests in `tests/test_cli.py` covering the `mailbox_save_callback` wiring, the `fail_on_output_error` interaction, and the `index_prefix_domain_map`/`email_results` fix.
- [x] Full suite: `pytest --cov tests/` — 727 passed.
- [x] `ruff check .` / `ruff format --check .` — clean.
- [x] `pyright` — 0 errors, 0 warnings.

Fixes #242.


seanthegeek:
I like the overall idea, but I have a concern about the retry semantics with multiple sinks.

The wiring is: `mailbox_save_callback` returns `not process_reports(batch)`, and `process_reports()` returns its accumulated error list — so if *any* sink fails, the whole batch stays in the mailbox, and on the next run/poll the messages are re-fetched, re-parsed (the dedup-cache update is deliberately skipped for failed batches so retry works at all), and `process_reports()` re-sends the batch to **every** configured sink, including the ones that already succeeded.

How bad the duplication is varies by sink:

- **Effectively safe on retry:** Elasticsearch, OpenSearch, and PostgreSQL dedupe via `AlreadySaved`, and S3 is idempotent too — the object key is deterministic (`.../year=/month=/day=/{report_id}.json` in `s3.py`), so a retry just overwrites the same object.
- **Duplicates on every retry:** Kafka, Splunk HEC, syslog, GELF, webhooks, the stdout JSON print, and the `--output` JSON/CSV files (`append_json`/`append_csv` will append the same rows again).

The PR body acknowledges the trade-off, but there are two distinct scenarios and the second is worse than the description suggests:

1. **Transient outage** (the #242 scenario): one or two duplicate deliveries to non-deduping sinks, then the batch archives. Arguably an acceptable price for not losing data — duplicates are recoverable, lost reports aren't.
2. **Persistent failure** (misconfigured Splunk token, decommissioned webhook, disk-full on `--output`): the batch *never* commits. In `--watch` mode, every check interval re-sends the entire — and growing, since new mail keeps arriving into the same folder — backlog to all the healthy non-idempotent sinks. That's unbounded duplication, the inbox never drains, and the only signal is a `logger.warning`. A single "poison" report that deterministically fails in one sink also pins its whole batch's other messages in this loop forever.

So the semantics are "at-least-once delivery to all sinks, all-or-nothing commit per batch" — a defensible design, but the failure mode in scenario 2 shifts from "lose data quietly" to "spam the working sinks and never drain the inbox." Requests, roughly in order of value-for-complexity:

- **A retry cap / quarantine path** would be the pragmatic middle ground: after N failed attempts, move the batch to a dedicated folder (a `Failed`/`Retry` sibling of the existing archive folders) instead of leaving it in `reports_folder` forever. That bounds the duplication and keeps the inbox draining while still never deleting data. Per-message attempt counting is awkward across IMAP/Graph/Gmail/Maildir, but even a process-local counter would cover watch mode. (Full per-sink commit tracking would be the complete fix, but that needs durable per-message/per-sink state and I don't think it's worth it for v1.)
- **At minimum:** escalate the retry log to `error` (it's the kind of thing people alert on), and document explicitly in `usage.md` that a *persistently* failing sink means unbounded re-delivery to Kafka/Splunk/syslog/GELF/webhook and repeated re-appends to `--output` files — the current changelog wording reads like a one-time duplication risk.

One smaller correction: since S3 is idempotent by key, the changelog's list of "destinations that don't deduplicate (Kafka, S3, Splunk HEC, Syslog, GELF, Webhook)" is slightly wrong about S3 — and it omits the `--output` file appends, which are arguably the most annoying duplicates since they silently corrupt CSVs people feed into other tooling.

---
*Analysis by [Claude Code](https://claude.com/claude-code) (Claude Fable 5), posted on behalf of @seanthegeek.*

seanthegeek:
Claude used the name quarantine, but I'm not crazy about it, but I'm not sure what else to call it either. In infosec terms, quarantines are where you put things that are likely malicious. I thought about "retry failures", but in DMARC, failure is also a type of report

Issue #472, opened 2024-02-15
Can msgraph be used to email the reports as it is done now with smtp?


PelaezLink:
Hello. After reading the documentation I implemented my solution using the MSGraph API with ClientSecret. I needed this as we use OAuth2 on our Exchange. Parsedmarc is pulling the reports correctly and sending them to elastic. The part I am missing is the sending of the email report. We used to have our own email server so we used smtp but now we need to send this from the same Exchange account is being read. Is this possible in any way? I didnt see such option in the .ini file. 
Thank you very much.

Kuzuto:
What kind of "email report" are you looking for it should do? I guess if you use Elastic, you have Kibana also. Is it overview/reports from here you want?

tomaskovacik:
https://github.com/ggpwnkthx/Microsoft-Graph-SMTP-Relay for example, there is a lot of options on GH

seanthegeek:
This shipped in parsedmarc 10.2.3: when `[smtp] host` is omitted and `[msgraph]` is configured, the summary email is now sent via the same authenticated Graph connection (`sendMail`, saved to Sent Items). Requires the Graph `Mail.Send` permission (`Mail.Send.Shared` for shared mailboxes under delegated auth) — see the "Sending the summary email via Microsoft Graph" docs section. Closing as implemented.

---
🤖 Triaged and closed by [Claude Code](https://claude.com/claude-code) at the maintainer's request.

PR #825, opened 2026-07-14
Send report summary via Microsoft Graph; make Graph failures observable

mkilijanek:
## Summary

Two related fixes:

**Send the periodic summary via Microsoft Graph** (tracks #472). Previously the summary email required `[smtp] host`, forcing M365 tenants that block legacy SMTP AUTH to stand up a separate SMTP relay just to send from the same mailbox they already read reports from.

**Make Microsoft Graph failures observable.** `cli.py` wrapped all Graph connection/usage errors in a bare `except Exception: logger.exception(...); exit(1)` that hid the actual Azure/Graph error, request ID, and which mailbox/tenant/auth method was involved.

### Send via Graph

When `[smtp] host` is omitted and `[msgraph]` is configured with a `[smtp] to` value, the summary is now sent through the same already-authenticated Graph mailbox connection (`MSGraphConnection.send_message()`, `/users/{mailbox}/sendMail`), saved to Sent Items. SMTP is preferred whenever `[smtp] host` is set, with no automatic fallback to Graph on SMTP failure -- reuses the existing connection, no new send-only config mode.

`[smtp] host`/`user`/`password`/`from` are now conditionally required -- only when `host` is present; `to` keeps its existing requirement either way. `email_results()`'s SMTP behavior and signature are unchanged; a new `email_results_via_msgraph()` shares its subject/message/zip-building logic via an extracted `_build_report_email_content()` helper, so both transports stay in lockstep.

`[smtp] from` has no effect on the Graph path -- Graph's `sendMail` always sends as the authenticated mailbox. Documented, along with the required `Mail.Send` permission and a caveat that delegated auth flows (`UsernamePassword`/`DeviceCode`) currently request only `Mail.ReadWrite`(`.Shared`), not `Mail.Send`, so app-only auth (`ClientSecret`/`Certificate`/`ClientAssertion`) is the supported path for sending today.

### Observable Graph failures

`MSGraphConnection` construction, mailbox fetch, message send, and `--watch` failures now catch `azure.core.exceptions.ClientAuthenticationError`, `kiota_abstractions.api_error.APIError` (the base of `ODataError`, the concrete type every Graph API non-2xx response raises), and `httpx.HTTPError` (unwrapped transport failures -- kiota never calls `raise_for_status()`) specifically, and log one clear ERROR line naming the mailbox, tenant, and auth method, plus the Graph `request-id`/`client-request-id` when the response exposes them (from the OData inner error, or the raw response headers as a fallback). Full traceback still preserved under `--debug`.

`--watch` previously had *no* Graph-specific error handling at all -- only `FileExistsError`/`ParserError` -- so a Graph auth, API, or transport error during a long-running watch (arguably the most likely real-world failure point, since token/certificate expiry surfaces there) crashed with a raw uncaught traceback. It now exits the same clean way as the other three call sites.

No new config options -- every field used (`opts.graph_mailbox`, `opts.graph_tenant_id`, `opts.graph_auth_method`) is already in scope.

## Test plan

- [x] New/updated tests in `tests/test_cli.py` (`TestMSGraphEmailResults`, `TestMSGraphFailureLogging`) covering: Graph-send success (payload content, not just "was called"), SMTP-preferred-when-both-configured (exact args), Graph-send-skipped-without-smtp-section, config validation, Graph-send failure logging, connection/fetch/watch failure logging with and without request-ids.
- [x] New tests in `tests/test_init.py` (`TestEmailResultsViaMsGraph`) proving `email_results_via_msgraph()` and `email_results()` build identical content for the same inputs.
- [x] `pytest --cov tests/` — 730 passed.
- [x] `ruff check .` / `ruff format --check .` — clean.
- [x] `pyright` — 0 errors, 0 warnings.
- [x] Docs build (`sphinx -b html`) clean.

Tracks #472.


seanthegeek:
Closing as already merged rather than rejected: the entire implementation here (code + tests) landed on `master` via #826 (31c928d), which was stacked on top of this branch — its squash-merge carried in all of this PR's commits, and its commit message even opens with this PR's title. `git diff master <this branch> -- parsedmarc/ tests/` is empty; the only delta left in this PR would revert the refreshed docs from #826, so there is nothing further to merge.

Thanks @mkilijanek for both halves of this work — the Graph summary-email sending and the Graph failure observability are live on `master` (credited to you via #826's merge commit) and will ship in the next release.

PR #826, opened 2026-07-14
Refresh Microsoft Graph docs: national clouds, examples, troubleshooting

mkilijanek:
## Summary

The `[msgraph]` docs were accurate but missed guidance the community has been asking for.

## Details

- `graph_url` now lists the actual national/sovereign-cloud endpoint values (GCC High, DoD, China/21Vianet -- verified against [Microsoft's own deployment docs](https://learn.microsoft.com/en-us/graph/deployments)), with an explicit warning that setting it alone is not sufficient -- the Entra ID auth endpoint isn't independently configurable in parsedmarc or mailsuite, so authentication always hits the global `login.microsoftonline.com` regardless of `graph_url`.
- A minimal working `[msgraph]` example for every auth method (`UsernamePassword`, `DeviceCode`, `ClientSecret`, `Certificate`, `ClientAssertion`) -- previously only `Certificate` had one, entangled with the SMTP-sending example.
- A reading-permission matrix alongside the existing sending one, so every auth method x own/shared-mailbox combination is explicit in one place for both directions.
- An accurate note on the `parsedmarc`-named token cache: it's a deliberate backward-compatibility choice from the 9.11.0 `mailsuite` extraction (mailsuite's own default cache name differs), not a migration users need to act on.
- A troubleshooting table for four error scenarios, verified against source rather than assumed:
  - "Admin consent required" and "Default folder Root not found" are still live and documented with real fixes.
  - `RuntimeError: Event loop is closed` (fixed in mailsuite 2.0.2) and the invalid-ISO-timestamp error (fixed in parsedmarc 9.5.1/9.5.5) are historical, already fixed below this project's dependency/version floor -- included for anyone searching the exact error text on an old install.

## Test plan

- [x] `sphinx -b html` docs build clean (only a pre-existing, unrelated warning on a different file).
- [x] No code changes.


seanthegeek:
One troubleshooting example that is missing is an expired ClientSecret. I can add that once I get to my desk. I'm just getting out of bed.

seanthegeek:
That error is pretty self explanatory, actually. So I'll merge this as is.

> The provided client secret keys for app 'APP ID' are expired. Visit the Azure portal to create new keys for your app: https://aka.ms/NewClientSecret, or consider using certificate credentials for added security: https://aka.ms/certCreds.

Issue #806, opened 2026-07-05
Mandatory `elasticsearch<7.14.0` dependency transitively pins urllib3<2 for all installs


PatrickV-code:
### Summary

`parsedmarc` lists `elasticsearch<7.14.0` and `elasticsearch-dsl==7.4.0` as **mandatory** dependencies in `pyproject.toml`. The `elasticsearch` 7.13.x client requires `urllib3 (<2,>=1.21.1)`, so every `parsedmarc` install is transitively pinned to the urllib3 **1.26.x** line — including for users who never use the Elasticsearch output backend.

### Why this matters

- urllib3 1.26.x is the final 1.x series; active maintenance and security fixes land in urllib3 **2.x**. Vulnerability scanners (e.g. Trivy) report HIGH-severity urllib3 advisories against 1.26.x that are only resolved in 2.x.
- Because the cap comes through a **mandatory** dependency, downstream projects that embed `parsedmarc` cannot move to urllib3 2.x without dropping `parsedmarc` or hand-editing its resolved dependency tree.
- The `<7.14.0` upper bound looks intentional — the elasticsearch 7.14+ client added a server "product check" that breaks against OpenSearch / non-matching clusters (cf. #452, #653) — so simply bumping the client version isn't a trivial change.

### Possible directions (either would resolve it)

1. **Make the Elasticsearch backend an optional extra**, mirroring the existing `postgresql` extra:
   ```toml
   [project.optional-dependencies]
   elasticsearch = [
       "elasticsearch<7.14.0",
       "elasticsearch-dsl==7.4.0",
   ]
   ```
   Users who need the ES output install `parsedmarc[elasticsearch]`; everyone else is freed from the transitive urllib3<2 cap. This is the lowest-risk option and matches the pattern already used for the optional PostgreSQL backend.

2. **Or** move to a modern `elasticsearch` / `elasticsearch-dsl` 8.x client line, which supports urllib3 2.x (with `elastic-transport` handling the product-check compatibility). This is a larger change but also frees urllib3 for ES users.

### Environment

- `parsedmarc` 10.2.0 (pins verified on `master` `pyproject.toml`)
- `elasticsearch` resolves to 7.13.4 → `urllib3 (<2,>=1.21.1)` → urllib3 1.26.x

Thanks for maintaining `parsedmarc` — the optional-extra approach in (1) looks small and low-risk, and I'd be glad to help if that direction is welcome.

seanthegeek:
[...]

PR #869, opened 2026-07-28
Cover index_prefix_domain_map and index_suffix in index migrations (#868)

### LINKED ISSUE #868: [Bug]: dkim_results_combined/spf_results_combined backfill skips indices named via index_prefix_domain_map
### parsedmarc version

10.4.0 (also present on `main` as of 5c7f4ba)

### Input backend

Maildir

### Environment

Docker (`ghcr.io/domainaware/parsedmarc:latest`) on Debian, OpenSearch 3 (`opensearchproject/opensearch:3`) + OpenSearch Dashboards 3, single-node. Multi-tenant deployment via [dmarc-msp](https://github.com/domainaware/dmarc-msp): one mailbox receives reports for many client domains, and each client's reports are routed to its own index prefix with `general.index_prefix_domain_map`.

### Sanitized config

```ini
[general]
save_aggregate = True
save_smtp_tls = True
index_prefix_domain_map = /etc/parsedmarc_domain_map.yaml

[opensearch]
hosts = https://opensearch:9200
user = admin
password = REDACTED
ssl = True
skip_certificate_verification = True
# NOTE: opensearch.index_prefix is deliberately NOT set — per-client
# prefixes come from index_prefix_domain_map instead.
```

`/etc/parsedmarc_domain_map.yaml`:

```yaml
acme_corp:
  - acme.example
  - acme-holdings.example
widgets_inc:
  - widgets.example
```

Resulting indices: `acme_corp_dmarc_aggregate-2026-07-27`, `widgets_inc_dmarc_aggregate-2026-07-27`, `acme_corp_smtp_tls-2026-07-27`, …

### Steps to reproduce

1. Run parsedmarc 9.x against OpenSearch with `general.index_prefix_domain_map` set and `opensearch.index_prefix` unset, so aggregate documents accumulate in per-client indices (`<prefix>_dmarc_aggregate-*`).
2. Upgrade to 10.4.0 and import the 10.4.0 dashboards.
3. Start parsedmarc and let the automatic `dkim_results_combined`/`spf_results_combined` backfill run.
4. Open the "DMARC aggregate" dashboard and look at the "SPF details" and "DKIM details" tables.

### Expected vs actual behavior

**Expected:** the backfill covers the indices this deployment actually writes to, so the auth-result detail tables show historical data after the upgrade.

**Actual:** the backfill silently does nothing, and the tables are unusable:

- "SPF details" shows the `missingBucket` label `none` in the `spf (scope / domain / result)` column for every row.
- "DKIM details" shows "No results found" (its `dkim_results_combined.keyword` bucket has `missingBucket: false`, so every document is dropped instead of labelled).

The cause is that `_init_output_clients()` derives the migration index names from `opts.opensearch_index_prefix` only, and never consults `index_prefix_domain_map` ([`cli.py:1611-1646`](https://github.com/domainaware/parsedmarc/blob/main/parsedmarc/cli.py#L1611-L1646)):

```python
os_aggregate_index = "dmarc_aggregate"
if opts.opensearch_index_prefix:          # unset in this deployment
    os_aggregate_index = f"{prefix}{os_aggregate_index}"
opensearch.migrate_indexes(aggregate_indexes=[os_aggregate_index], ...)
```

`migrate_indexes()` then runs its guard `count` against `f"{name}*"` → `dmarc_aggregate*`, which matches **none** of the real `<prefix>_dmarc_aggregate-*` indices. Because the guard passes `allow_no_indices=True` and `ignore_unavailable=True`, a zero-match wildcard returns `count: 0`, which is indistinguishable from "already backfilled" — so the migration is skipped with no log line at any level.

Same failure for `policies_combined`/`failure_details_combined` on the SMTP TLS dashboard, via the `smtp_tls*` pattern.

This is a mismatch between the migration path and the save path. The save path honors both sources — `index_prefix=opts.opensearch_index_prefix or get_index_prefix(report)` ([`cli.py:1862`](https://github.com/domainaware/parsedmarc/blob/main/parsedmarc/cli.py#L1862)) — while the migration path honors only the first. Elasticsearch is affected identically (`cli.py:1566-1599`).

The manual `_update_by_query` from `docs/source/elasticsearch.md` does repair the data if the index pattern is widened to `*dmarc_aggregate*`, so this is purely about which indices the automatic backfill targets.

### Logs or traceback

None — the absence of output is part of the bug. With `--debug`, the backfill produces no line for the skipped per-client indices; `migrate_indexes()` only logs when `count > 0` (`INFO`) or when the request raises (`WARNING`), and a wildcard matching zero indices does neither.

### Sample report availability

Not needed to reproduce — any aggregate report indexed by a pre-10.4.0 release under an `index_prefix_domain_map` prefix reproduces it. Happy to share a sanitized aggregate report or an anonymized `_mapping`/document dump if useful.

### Proposed fix

Call-site only; `migrate_indexes()` already accepts a list and loops one pattern per name, so it needs no change.

1. Extract the `index_prefix_domain_map` key normalization from the `get_index_prefix()` closure (`cli.py:1730-1736`) into a module-level `_normalize_index_prefix()`, and have both the write path and the migration path use it, so they cannot drift.
2. Add a `_migration_index_names(base_name, configured_prefix, index_prefix_domain_map)` helper that mirrors the save-time precedence: a configured `index_prefix` wins outright and suppresses fan-out (otherwise a deployment that sets one could submit `_update_by_query` against index patterns it never writes, which on a shared cluster is another deployment's data); with no configured prefix, return the unprefixed base name **plus** one name per normalized map key, deduplicated. The unprefixed name has to stay, because aggregate and failure reports for a domain absent from the map are still saved with no prefix — only SMTP TLS reports are filtered out (`filter_smtp_tls_reports_for_index_prefix`).
3. Give `_init_output_clients()` an `index_prefix_domain_map=None` keyword and pass it at both call sites (`cli.py:2509` and the reload path at `cli.py:3063`, which already has `new_index_prefix_domain_map` in hand one line earlier). The reload path then picks up newly onboarded prefixes on SIGHUP for free, and the `count` guard keeps that idempotent.
4. Log the resolved index-name list at `DEBUG` before the loop. A silent skip is what let this ship; a fresh install legitimately has no indices, so `WARNING` would be noise, but the operator should be able to see what the backfill decided to target.

Cost is one `count` per name per report type at startup, each cheap and short-circuiting on a non-existent pattern.

I can open a PR with this plus tests: `_migration_index_names` units (configured prefix wins and suppresses fan-out, map fan-out includes the unprefixed base, dedup of keys that normalize identically, suffix-before-prefix ordering, `None` map preserves current behavior) and a `test_cli.py` assertion on the `aggregate_indexes=` kwarg when a domain map is configured, since those tests already patch `migrate_indexes`.

seanthegeek:
Fixes #868. Also bumps the version to 10.4.1 and renames the CHANGELOG heading, since this is intended as the 10.4.1 patch release.

## The bug

`_init_output_clients()` built the Elasticsearch/OpenSearch migration index names from `opts.{elasticsearch,opensearch}_index_prefix`/`_index_suffix` alone, while the save path also honors `general.index_prefix_domain_map` (`index_prefix=opts.*_index_prefix or get_index_prefix(report)`).

In a multi-tenant deployment — per-tenant prefixes from the map, no `index_prefix` set, exactly as the multi-tenant docs instruct — the backfill guard `count` therefore ran against `dmarc_aggregate*`/`smtp_tls*`, patterns matching none of the real `<tenant>_dmarc_aggregate-*` indexes. Since the query passes `allow_no_indices=True`, a zero-match wildcard returns `count: 0`, which is indistinguishable from "already backfilled" — so the `dkim_results_combined`/`spf_results_combined` and `policies_combined`/`failure_details_combined` backfills were skipped **silently, with no log line at any level**. The 10.4.0 dashboards' "DKIM details" table shows "No results found" and every "SPF details" row shows `none`.

The same silent skip has a second trigger the issue didn't mention: adding or changing `index_suffix` leaves the operator's own earlier `dmarc_aggregate-*` indexes unmatched by `dmarc_aggregate_prod*`.

## The fix

A new `_migration_index_names(base_name, index_suffix, configured_prefix, index_prefix_domain_map)` resolves the target names, widening both axes:

- **Prefix.** No configured `index_prefix` → the unprefixed name plus one per tenant prefix in the map, deduplicated. The unprefixed name stays because aggregate and failure reports for a domain absent from the map are still saved without a prefix, and every report type may have indexes predating the map. A configured `index_prefix` wins outright and suppresses the fan-out, matching save-time precedence, so such a deployment never submits an `_update_by_query` against a pattern it does not write to.
- **Suffix.** Both the suffixed and unsuffixed names, suffixed first. Trade-off noted in the docstring and CHANGELOG: because callers turn each name into an `f"{name}*"` pattern, the unsuffixed pattern is a strict superset, so on a shared cluster it also matches other deployments' suffixes, and both patterns can submit an overlapping (idempotent, `conflicts=proceed`) `update_by_query` on first run.

`get_index_prefix()` and the migration path now share `_normalize_index_prefix()`, so the two cannot drift. The resolved lists are logged at `DEBUG` — a silent skip is what let this ship. The SIGHUP reload path passes the freshly parsed map, so a newly onboarded tenant is covered without a restart; the `count` guard keeps that idempotent.

## Also: the legacy `published_policy.fo` migration

Investigating "any migration, not just the backfill" turned up that this one has been unable to complete since mapping types were removed in Elasticsearch 7 (and they never existed in OpenSearch):

- it read the field mapping in the type-keyed response shape, so `if doc not in fo_mapping: continue` always fired; and
- `put_mapping(doc_type=...)` would have raised `TypeError` on either current client — dead code hiding a crash, which is why tests mocking the old response shape passed.

It now reads either shape, uses each client's current `put_mapping()` signature (`properties=` for elasticsearch-py 8.x, `body=` for opensearch-py), and takes its names in a separate `legacy_fo_indexes` argument. Those are exact names, not patterns: 5.0.0 introduced date-suffixed index names in the same release that fixed the `fo` declaration, so an affected index has no date component — but it can be prefixed or suffixed, since both options date back to 4.1.0, so the configured `index_prefix`/`index_suffix` are applied. Tenant prefixes deliberately are not: `index_prefix_domain_map` arrived in 8.19.0, and this migration renames the index it rebuilds. The Elasticsearch copy, removed as unreachable during the #806 client migration, is restored now that the cause is understood.

## Also: `index_prefix_domain_map` shape validation

The YAML was used unvalidated. A non-string key raised mid-save; a scalar value silently matched by substring, since `in` on a `str` is a substring test — `"example.co" in "example.com"` is `True` ([Python docs](https://docs.python.org/3/reference/expressions.html#membership-test-operations)). Both are now rejected at startup with a `ConfigurationError`; an empty file still means "unset". The check sits in `_parse_config`, so it covers the SIGHUP reload too.

## Verification

- `ruff check .`, `ruff format --check .`, `pyright` — all clean repo-wide from the repo root.
- `pytest --cov` — 918 passed. `elastic.py` and `opensearch.py` at 100%; `cli.py` unchanged at 81%, and none of its missing lines are new.
- Eight of the eleven new tests fail against the unfixed library (verified by stashing `parsedmarc/`); the three that pass beforehand are the negative-half and preserved-behavior cases, and say so in their docstrings.
- Docs build resolves the new cross-reference, adding no warnings.
- End-to-end CLI run against the working tree with `index_suffix` + a domain map prints the expected resolution, including `legacy_fo=['dmarc_aggregate_prod', 'dmarc_aggregate']`.

A cold review of the final diff caught a real defect in the first draft: it asserted that `index_prefix`/`index_suffix` postdated the `fo` mapping fix, so the legacy migration only needed the bare `dmarc_aggregate`. Commit a7fb207 (4.1.0) added both options while `fo = Integer()` was still in force through 4.5.0, so that draft would have missed prefixed and suffixed legacy indexes — and removed coverage the pre-existing OpenSearch code happened to have. Fixed here, along with three other incorrect historical claims that had been repeated across the code, tests and CHANGELOG.


seanthegeek:
Both review points from the last round are addressed in 5aaff57, verified against live Elasticsearch 8.19.0 and OpenSearch 3 containers rather than mocks — the migration is exactly the kind of code where a mocked test can't answer the question.

## Retry safety after an interrupted attempt — confirmed, fixed

Reproduced on a live cluster. Seeding an original index with `published_policy.fo` mapped as `long`, plus a leftover `dmarc_aggregate-v2` from a notional earlier attempt, the unfixed code gives:

```
[FAIL] retry deleted the original
[FAIL] retry produced the corrected mapping -- got None
[FAIL] stale debris discarded, real document present -- sources = [{'junk': 'from an interrupted run'}]
```

`create()` raises `resource_already_exists_exception` inside the same `try` that swallows migration errors, so the original is never migrated and the debris survives — on that startup and every one after it.

A leftover target is now discarded first. That is safe precisely because reaching that point means the original still holds the data: it is deleted only after the reindex succeeds. Both suites gained a regression test, each confirmed to fail against the unfixed code.

## Dotted field name under `properties` — not reproducible

This one does not hold. Putting both mapping bodies to each cluster:

```
dotted-key:    ACCEPTED -> top-level keys=['published_policy'],
               published_policy.fo={'type': 'text', 'fields': {'keyword': {...}}}
nested-object: ACCEPTED -> top-level keys=['published_policy'],
               published_policy.fo={'type': 'text', 'fields': {'keyword': {...}}}
```

Both are accepted and produce a byte-identical mapping — the dot is expanded into `published_policy` → `properties` → `fo`, matching the `subobjects` documentation ("Elasticsearch accepts fields that contain dots in their names, which get expanded to their corresponding object structure"). Nothing was at risk of failing at runtime, and the field is not created as a literal dotted leaf.

I switched to the nested form regardless, since dot expansion is conditional on the object's `subobjects` setting while the nested shape never is, and derived the object/leaf names from the existing dotted constant so the mapping written can't drift from the field the type check reads. But the change is robustness, not a bug fix, and the comment in the code says so rather than implying the old form was broken.

## End-to-end coverage

The full migration now has a live-cluster check on both engines: seed the pre-5.0.0 `long` mapping, run `migrate_indexes()`, and assert the `-v2` index is created with `fo` mapped `text` plus its `keyword` sub-field nested under `published_policy`, the document is carried over with its content intact, the original is deleted, and the interrupted-attempt path recovers. All pass on both.

That also exposed a weakness in the existing mocked tests: a single shared `Index` mock cannot express "the original exists but its migration target does not", which is the ordinary case and the one the retry fix turns on. Both suites now build per-name mocks, and the happy path additionally asserts the target index is never deleted.

Suite: 921 passed, `elastic.py` and `opensearch.py` at 100%, `ruff` and `pyright` clean.

PR #872, opened 2026-08-13
Fixed Grafana JSON parameter errors

ericericsw:
## Summary
Correcting the incorrect interval name: fixed_interval->interval
Maintaining the fixed_interval setting will cause errors in the dashboard output.
- 

## Why
image
image
image

- 

## Testing
image
- 

none

### DIFF
--- a/dashboards/grafana/Grafana-DMARC_Reports.json
+++ b/dashboards/grafana/Grafana-DMARC_Reports.json
@@ -156,7 +156,7 @@
         "y": 2
       },
       "id": 6,
-      "fixed_interval": null,
+      "interval": null,
       "legend": {
         "percentage": false,
         "show": true,
@@ -191,7 +191,7 @@
               "field": "date_begin",
               "id": "2",
               "settings": {
-                "fixed_interval": "auto",
+                "interval": "auto",
                 "min_doc_count": 0,
                 "trimEdges": 0
               },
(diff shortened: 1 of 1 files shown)


seanthegeek:
Thanks for the fix — confirmed correct, and worth documenting how this key ended up in the dashboard JSON.

**What went wrong:** Grafana's Elasticsearch datasource has never used `fixed_interval` in its dashboard JSON model. The `date_histogram` bucket aggregation settings expect `interval` (see `DateHistogramSettings` in Grafana's `dataquery.gen.ts`: `interval`, `min_doc_count`, `offset`, `timeZone`, `trimEdges`), and Grafana itself translates that into `fixed_interval`/`calendar_interval` when it builds the query DSL it sends to Elasticsearch (`pkg/tsdb/elasticsearch/client/models.go`). With `fixed_interval` in the JSON, current Grafana silently ignores the setting, which is what produced the broken panels in the screenshots above.

**When it went wrong:** The key was renamed in #366 (December 2022) as a workaround for Elasticsearch 8 removing `interval` from `date_histogram` (grafana/grafana#19911). That workaround made sense with the Grafana versions of that era, which passed settings through to Elasticsearch more literally — but once Grafana started handling the ES 8 translation internally, the renamed key stopped matching anything in the model and the workaround became the bug. This PR restores the dashboard to what current Grafana expects.

I also verified the PR head parses as valid JSON with zero remaining `fixed_interval` occurrences, and that the panel-level `interval` keys this reactivates (e.g. `$interval` on the messages stat panel) match the dashboard's template variables. The array reflows (`calcs`, `tags`, `time_options`) are cosmetic only.

Issue #883, opened 2026-08-27
[Feature]: Make the output and mailbox integrations optional extras


rokokol:
### Problem statement

parsedmarc's parsing core — `parse_report_file` / `parse_aggregate_report_xml` / `parse_smtp_tls_report_json` — needs none of the delivery integrations, yet `pyproject.toml` declares Elasticsearch, OpenSearch, Kafka, boto3, Azure (identity + monitor) and `mailsuite[gmail,msgraph]` as hard dependencies. Installed, that is roughly a gigabyte of site-packages (measured on 10.4.3: ~1070 MiB; ~300 MiB with the integrations removed).

The use case this hurts: running parsedmarc **as a library on the mail host itself** — a small self-hosted node parses the DMARC and TLS-RPT reports that arrive in its own mailbox and stores rows locally, shipping nothing to Elasticsearch, Splunk or any cloud mailbox API. On a small VPS disk the unused integration SDKs dominate the install.

### Proposed behavior

Staged, so each step stays reviewable:

1. Import the optional mailbox connections lazily in `parsedmarc.mail` — mailsuite already guards them behind a lazy `__getattr__` with a clear "requires the extra" ImportError, and the eager re-export currently defeats that guard, making `import parsedmarc` itself fail without the gmail/msgraph stacks. (PR to follow.)
2. Guard the output-module imports in `cli.py` (`elastic`, `opensearch`, `kafkaclient`, `splunk`, `gelf`, `loganalytics`) the same way, erroring helpfully only when the corresponding config section is actually used.
3. Move the integration dependencies into `[project.optional-dependencies]` (`elastic`, `opensearch`, `kafka`, `s3`, `gelf`, `msgraph`, `gmail`, …) with an umbrella `all` extra, and point the CLI install docs at `parsedmarc[all]` so current CLI users see no change.

Happy to send the follow-up PRs if the direction is acceptable.

### Alternatives considered

Vendoring a hand-written parser for the two report formats — discarded, since parsedmarc's format detection and robustness against real-world reporter quirks is exactly the value of using the library. Distro-side dependency stripping works (that is how the measurement above was made) but has to patch the eager re-export, which is better fixed once upstream.

### Compatibility or operational impact

`pip install parsedmarc` keeping all extras by default (or docs pointing at `[all]`) keeps the CLI experience unchanged; only consumers who explicitly install a slimmer set see a difference.

PR #884, opened 2026-08-27
Import optional mailbox connections lazily in parsedmarc.mail

rokokol:
## Summary

- `parsedmarc.mail` re-exported the optional Gmail and Microsoft Graph connections eagerly, so `import parsedmarc` failed with mailsuite's "requires the extra" ImportError whenever the package is installed without those stacks (distro/Nix packaging, `pip install --no-deps`, vendored environments)
- The optional imports now sit under `TYPE_CHECKING` for the type checker and behind `try`/`except ImportError` at runtime; a missing extra yields a placeholder class that raises the original ImportError at construction, so the failure moves from import time to actual use, with the same message

## Why

- mailsuite ≥ 2.0 deliberately guards these connections behind a lazy `__getattr__`; the eager re-export here defeats that guard
- The parsing core (`parse_report_file` and friends) needs none of the mailbox stacks — the motivating case is running parsedmarc as a library on the mail host itself, parsing reports from the local mailbox with nothing shipped to cloud mailbox APIs
- First step of #883

## Testing

- `ruff check` and `ruff format --check` clean on the file
- `pyright parsedmarc/mail/__init__.py`: 0 errors, 0 warnings
- `GITHUB_ACTIONS=true pytest tests/`: 860 passed, 73 subtests passed; the one failure (`tests/test_utils.py::TestTimestampAssumeUtc::testWithoutAssumeUtcNaiveIsLocal`) reproduces identically on an unmodified checkout in this environment (UTC+3 host), so it is unrelated
- With the gmail/msgraph stacks absent: `import parsedmarc` succeeds, `GmailConnection()` raises the extra's original ImportError, `isinstance` checks against the placeholders are well-formed and return False

- No behavior change for installations that have all dependencies: the real classes import exactly as before
- `AuthMethod` becomes `None` when the msgraph extra is missing — reached only by code paths that already could not run without `MSGraphConnection`

## Related Issue

- First step of #883

## Checklist

- [x] Tests added or updated if behavior changed — no behavior change with full dependencies; the guarded path is exercised by the checks above
- [ ] Docs updated if config or user-facing behavior changed — not user-facing until #883 moves the dependencies into extras

### DIFF
--- a/parsedmarc/mail/__init__.py
+++ b/parsedmarc/mail/__init__.py
@@ -5,16 +5,49 @@
 The implementations live in :mod:`mailsuite.mailbox` (extracted from
 parsedmarc in mailsuite 2.0.0). This module re-exports them so
 ``parsedmarc.mail`` remains a stable import path for downstream consumers.
+
+The Gmail and Microsoft Graph connections require optional extras. Importing
+this module must not: when an extra is missing, its connection class is
+replaced by a placeholder that raises the same ImportError at construction
+time, so ``import parsedmarc`` works for consumers that only parse reports.
 """
 
+from typing import TYPE_CHECKING
+
 from mailsuite.mailbox import (
-    GmailConnection,
     IMAPConnection,
     MailboxConnection,
     MaildirConnection,
-    MSGraphConnection,
 )
-from mailsuite.mailbox.graph import AuthMethod
+
+if TYPE_CHECKING:
+    from mailsuite.mailbox import GmailConnection, MSGraphConnection
+    from mailsuite.mailbox.graph import AuthMethod
+else:
+    try:
+        from mailsuite.mailbox import GmailConnection
+    except ImportError as error:
+        _gmail_import_error = error
+
+        class GmailConnection:
+            """Placeholder raising the gmail extra's ImportError when used."""
+
+            def __init__(self, *args, **kwargs):
+                raise _gmail_import_error
+
+    try:
+        from mailsuite.mailbox import MSGraphConnection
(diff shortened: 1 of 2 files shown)


seanthegeek:
Thanks @rokokol — the approach here is exactly right, and I've verified it end-to-end (including simulating the missing extras with an import blocker). I pushed a maintainer commit (046eb6c) with three follow-ups from review rather than round-tripping them:

- **Added `tests/test_mail.py`.** CI always has the extras installed, so the new `except ImportError` branches would never execute there, and this repo's codecov patch gate is blocking. The tests use a `sys.meta_path` blocker to simulate the absent stacks and cover both guards, the placeholder behavior, and `isinstance` well-formedness, with careful `sys.modules` purge/restore so nothing leaks into other suites.
- **`AuthMethod = None` → a placeholder object.** `None` turned any use into a cryptic `AttributeError: 'NoneType' ...`/`TypeError` — e.g. `parsedmarc.cli`'s `AuthMethod.UsernamePassword.name` — instead of the actionable "pip install mailsuite[msgraph]" message. The placeholder raises that ImportError on non-dunder attribute access, iteration, call, and subscript, covering every Enum usage shape.
- **Placeholders now raise a fresh `ImportError` chained `from` the stored original**, so repeated constructions don't keep appending traceback frames to one shared exception object.

No changelog entry: with all dependencies still hard-required this isn't a user-facing fix yet — it's groundwork for the #883 follow-up that introduces the extras.

*This comment and the review behind it were written by Claude (Fable 5), working under my direction.*

rokokol:
Thanks @seanthegeek !

PR #898, opened 2026-09-09
Use sys.exit() instead of the site-injected exit() builtin

seanthegeek:
Replaces the bare `exit(...)` builtin with `sys.exit(...)` across the CLI and the maintainer map-tooling scripts, and removes four dead `index_date` stores. Found by a local CodeQL `python-code-quality.qls` scan (`py/use-of-exit-or-quit`, `py/multiple-definition`) — note the repo's CodeQL default setup does not run the code-quality suite, so these never appeared as alerts.

## Why `exit()` is wrong here

`exit` and `quit` are installed into `builtins` by the `site` module — for every run that imports `site`, which is why bare `exit()` worked in these scripts at all. Under `python -S`, or an embedded interpreter that skips `site`, the name is absent. The failure paths that called it would then raise `NameError` instead of exiting, so the breakage lands exactly where a user is already troubleshooting a bad config or an unreachable mailbox.

29 call sites: 20 in `cli.py`, 5 in `find_unknown_base_reverse_dns.py`, 4 in `sortlists.py` (which needed `import sys` added). The three signal handlers are deliberately untouched — they correctly use `os._exit`, and the comment explaining why `sys.exit` is unsafe there still holds.

## Dead `index_date` stores

`save_aggregate_report_to_elasticsearch` and `save_aggregate_report_to_opensearch` each computed `index_date` from the report-level `metadata["begin_date"]`, then recomputed it inside the per-record loop from `record["interval_begin"]` before the only read. The outer value could never reach an index name — it was not merely redundant but computing a *different* date, dead since the per-record recomputation arrived in 9.0.0 (#629). Both branches of the surviving `if monthly_indexes:` still assign, the read sits inside the same loop, and an empty `records` list reaches neither. **`monthly_indexes` behavior is unchanged.**

## One adjacent bug fixed

`find_unknown_base_reverse_dns.py:199` exited with status **0** immediately after printing a duplicate-entry error, while the identical check at :177 exits 1. In a shell `&&` chain that error read as success. This is a line the diff already rewrote.

## Found but deliberately not fixed

The same script prints `Error: ... does not exist` at ~:170 and ~:190 and then falls through into an `open()` that crashes with a traceback instead of exiting cleanly. Pre-existing, not on lines this PR touches — left for a separate change.

## Verification

`ruff check .`, `ruff format --check .`, `pyright` (0 errors, 0 warnings), and `GITHUB_ACTIONS=true pytest tests/` (989 passed) all clean from the repo root. No test patched `exit`; the `assertRaises(SystemExit)` tests are agnostic to which form raised it. Exit statuses are unchanged.


seanthegeek:
`codecov/patch` fails here at 30% (14 uncovered lines in `cli.py`). Those 14 lines are not new gaps — they are the same statements that were already uncovered before this PR, rewritten from `exit(...)` to `sys.exit(...)`. This is the "rewritten lines are new patch lines even when behavior is intentionally identical" case in AGENTS.md.

Verified rather than asserted. Full suite run with `--cov=parsedmarc --cov-report=term-missing` on `origin/master` and on this branch:

```
master  parsedmarc/cli.py   1528 statements   276 missing   82%
branch  parsedmarc/cli.py   1528 statements   276 missing   82%
```

Then comparing the two uncovered *statement* sets by source text rather than line number (line numbers shift), normalising the intended `exit(` → `sys.exit(` rewrite:

```
missing count  master: 373   branch: 373
uncovered statement sets identical after normalising exit->sys.exit: True
uncovered sys.exit lines in branch: 14
```

So no statement went from covered to uncovered. Codecov's own diff agrees from the other direction: misses are unchanged at 346, and hits drop by exactly 6 — the deleted dead `index_date` stores, which *were* covered.

Covering those 14 would mean testing `_main()`'s deep error paths (MS Graph and Gmail connection failures, config errors), which is the territory AGENTS.md explicitly calls out as not worth faking: PR-B (#775) halted `cli.py` at 69% for this reason. I'd rather leave the gate red with this explanation than add mocks that verify themselves.

*— Claude Code (Claude Opus 5), on behalf of @seanthegeek*

PR #907, opened 2026-09-11
docs: add CONTRIBUTING.md with dev setup and test instructions

maksimtech:
## Summary

- Add CONTRIBUTING.md documenting how to set up the development environment and run tests

## Why

- Without `pip install -e ".[build,all,dev]"`, three tests fail due to missing `pytz` and `mailsuite[gmail]` dependencies, which is confusing for new contributors
- Tests that perform real DNS lookups hang indefinitely in local environments unless `GITHUB_ACTIONS=true` is set — this is not documented anywhere
## Testing

- Verified all 744 tests pass with `GITHUB_ACTIONS=true pytest tests/` after following the documented setup

- Documentation only — no code changes, zero risk

## Related Issue

- N/A

## Checklist

- [x] Docs updated if config or user-facing behavior changed


seanthegeek:
Thanks for taking the time to open this, and for the clear write-up of what tripped you up.

I'm going to close this one, though, because the repository already has a `CONTRIBUTING.md`, and this PR replaces it rather than adding to it. Both of the problems in the "Why" section are covered by the existing file:

- The missing `pytz` and `mailsuite[gmail]` come from the `build` and `all` extras. The existing guide's install line, `pip install ".[build,all,postgresql]"`, pulls both in. (There's no `dev` extra in `pyproject.toml`, so `.[build,all,dev]` would silently install less than intended.)
- `GITHUB_ACTIONS=true` for skipping live DNS lookups is already documented in the "To skip DNS lookups during tests" section.

The replacement would also drop a fair amount of material we rely on: the docs build step, the sample-report smoke run, the PR guidelines, the rebase workflow, and the notes on reading Codecov. And it changes the test and formatting commands so they no longer match what CI enforces (`pytest --cov`, `ruff format --check`).

If there's a specific part of the existing guide that was unclear when you first set up, I'd genuinely like to hear it. A small PR that fixes that wording would be very welcome.

---
*Drafted with the help of Claude Code, reviewed and posted by the maintainer.*

